PR #102, opened 2014-07-21
Parser: Add support for multi-document tag directives

ztzg:
The spec at <http://yaml.org/spec/1.1/#l-first-document> says:
    
> If the document specifies no directives, it is parsed using the same settings as the previous document.

Without this change, `ParseDocumentEnd()` unconditionally clears the collection of tag directives, which causes the following `SemanticError`:

> While parsing a node, find undefined tag handle.

when parsing a multi-document stream such as:

    %YAML 1.1
    %TAG !x! tag:example.com,2014:
    --- !x!foo
    x: 0
    --- !x!bar
    x: 1

where both instances of `!x!` stand for for `tag:example.com,2014`.

The main patch postpones the clearing of `tagDirectives` until a new set of directives is encountered at the start of a document.

### DIFF
--- a/YamlDotNet.Test/Core/ParserTests.cs
+++ b/YamlDotNet.Test/Core/ParserTests.cs
@@ -310,6 +310,29 @@ namespace YamlDotNet.Test.Core
 				StreamEnd);
 		}
 
+		[Fact]
+		public void VerifyTokenWithMultiDocTag()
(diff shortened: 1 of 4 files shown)


aaubry:
This has been resubmitted as #142 and it is already merged.

PR #104, opened 2014-08-13
Added Xamarin Android class library.

felixcollins:
I did try to create a PCL that supported Android and iOS but ran into loads of problems. At least it was really easy to just compile it for Xamarin.Android. Is it just me or is the whole idea of mulitple flavours of PCL a bit broken?


aaubry:
YamlDotNet now builds for multiple targets, including portable, so I'm closing this.

PR #110, opened 2014-11-20
Apply naming convention to keys in dynamic dict

mattmiller85:
When serializing ExpandoObjects, the naming convention was not getting
applied to the keys like it does for statically typed objects.


aaubry:
I'm sorry, I did not have time to review this pull request earlier. I think that this change would make the naming convention be applied to dictionaries other than ExpandoObject, right? If that is the case, this could cause unintended behavior. I think that if one is serializing a Dictionary<string, T>, the keys should be serialized as-is, regardless of the naming convention.

mattmiller85:
You are correct, but only if the dictionary implements the generic IDictionary, and not the non-generic IDictionary. All non-generic IDictionaries get sucked up by the first if() in TraverseObject in FullObjectGraphTraversalStrategy and don't go into TraverseGenericDictionary. I tried to write a test for this case, but I can't test if the IDictionary is an ExpandoObject since YamlDotNet is targeting .NET 3.5.

I'm not sure where to go from here

mattmiller85:
Literally right after I typed that I thought of a solution using reflection to test for ExpandoObject. Added a test to show it.

aaubry:
Thanks for this. I only changed the test for the ExpandoObject type to use the full type name instead of using Contains, and I also made the formatting compliant with the project (tabs instead of spaces :) )

PR #115, opened 2014-12-14
Address issue 114

jeffman:
At least a partial fix for issue 114. Not fully sure yet if other things need to be fixed as well.


aaubry:
This should have been fixed by #179.

PR #116, opened 2014-12-26
Anchor assignment during YamlStream.Save(): ability to opt out

jeffman:
I noticed that assigning node anchors can be extremely slow for large `YamlDocument`s; for example, writing a ~5.3 MB YAML file takes about 28 minutes, and the vast majority of that time is taken from assigning node anchors. 

If I know ahead of time that my document does not contain any cross-references, then I can safely skip the anchor assignment and the output will be unaffected. This brought the writing time down from 28 minutes to 4 seconds for that 5.3 MB file.

This change adds an overload to `YamlStream.Save()` which allows the user to opt out of anchor assignment. This defaults to `true` (which is the current behaviour and which makes this a non-breaking change), but it can be changed to `false` to skip anchor assignment and consequently save lots of time.

The user would only need to be aware that explicitly using the overload with `false` would cause cross-referenced nodes to be emitted independently, which may or may not be desirable based on their requirements.

### DIFF
--- a/YamlDotNet/RepresentationModel/YamlDocument.cs
+++ b/YamlDotNet/RepresentationModel/YamlDocument.cs
@@ -169,14 +169,15 @@ namespace YamlDotNet.RepresentationModel
 			visitor.AssignAnchors(this);
 		}
 
-		internal void Save(IEmitter emitter)
-		{
-			AssignAnchors();
+        internal void Save(IEmitter emitter, bool assignAnchors = true)
+        {
+            if (assignAnchors)
+                AssignAnchors();
 
-			emitter.Emit(new DocumentStart());
-			RootNode.Save(emitter, new EmitterState());
-			emitter.Emit(new DocumentEnd(false));
-		}
+            emitter.Emit(new DocumentStart());
+            RootNode.Save(emitter, new EmitterState());
+            emitter.Emit(new DocumentEnd(false));
+        }
 
 		/// <summary>
 		/// Accepts the specified visitor by calling the appropriate Visit method on it.
--- a/YamlDotNet/RepresentationModel/YamlDocument.cs
+++ b/YamlDotNet/RepresentationModel/YamlDocument.cs
@@ -169,15 +169,15 @@ namespace YamlDotNet.RepresentationModel
 			visitor.AssignAnchors(this);
 		}
 
-        internal void Save(IEmitter emitter, bool assignAnchors = true)
-        {
-            if (assignAnchors)
-                AssignAnchors();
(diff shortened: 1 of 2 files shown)


jeffman:
The bottleneck is the dictionary lookup. I just tried writing an alternate `AnchorAssigningVisitor` that doesn't use a dictionary, and I saw the same improvements (from 28 minutes down to about 4.2 seconds). The consequence is that I have to add an internal `Visited` property to `YamlNode` to facilitate the visitation state -- basically I'm recording it in the node directly, rather than indirectly using a dictionary.

Looks something like this:

```csharp
private class AnchorAssigningVisitor2 : YamlVisitor
{
	private readonly HashSet<string> existingAnchors = new HashSet<string>();
	private Random random = new Random();

	public void AssignAnchors(YamlDocument document)
	{
		existingAnchors.Clear();

		var stateVisitor = new VisitStateVisitor();
		stateVisitor.AssignState(document, false); // just resets each node.Visited state to false

		document.Accept(this);
	}

	private void VisitNode(YamlNode node)
	{
		if (string.IsNullOrEmpty(node.Anchor))
		{
			if (node.Visited)
			{
				string anchor;
				do
				{
					anchor = random.Next().ToString(CultureInfo.InvariantCulture);
				} while (existingAnchors.Contains(anchor));
				existingAnchors.Add(anchor);

				node.Anchor = anchor;
			}
			else
			{
				node.Visited = true;
			}
		}
		else
		{
			existingAnchors.Add(node.Anchor);
		}
	}
	...
}
```

Would this be a better solution? Although I agree that it would still be nice to yield cross-referencing control to the user.

aaubry:
I'm sorry, I could not find time to answer earlier. I agree with @rogernorling that the correct solution is that it should not take so long. I also agree that instead of comparing by value, we should simply compare by reference and the programmer be responsible for using the same reference if he wants.
It should be a matter of removing the comparer from https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/RepresentationModel/YamlDocument.cs#L101.

@jeffman , do you want to fix this yourself, or would you rather let me attempt to fix it?

ps. @rogernorling, good to hear from you! It's all right, I haven't had much time to dedicate to this project either.

aaubry:
I have decided to merge this PR anyways because it seems reasonable to be able to opt out from the anchor assignment. I also [removed the "by value" comparer](https://github.com/aaubry/YamlDotNet/commit/5157bc2858cb70ed89e3951513650787532e6b3e), so the anchor assignment process should be quick anyways.

PR #118, opened 2015-01-30
Update DefaultObjectFactory.cs

paranoidandroid522:
Adds support for deserializing objects with non-public default ctor (which in some cases allows better encapsulation)


paranoidandroid522:
I have to admit that this comes with a performance cost, so it might be better to make this behavior configurable. It serves me well enough in my project, though.

paranoidandroid522:
Found a bug - pull request withdrawn.

PR #119, opened 2015-01-30
Support for non-public ctor in DefaultObjectFactory (attempt no. 2 ;-))

paranoidandroid522:


aaubry:
Are you sure this is a good idea?
The DefaultObjectFactory is a simple and default implementation of `IObjectFactory`. I don't think that it should be able to call private constructors.
It is easy to replace the implementation of `IObjectFactory` with another. Therefore, I think if an application has special needs regarding object instantiation, it should provide its own implementation of the interface.

paranoidandroid522:
Well, I am kind of used to serializers (like JSON.NET or the WCF DataContractSerializer) to invoke private ctors by default. This facilitates better encapsulation in some situations.
The only possible issue I see here is a slight performance penalty through reflection - one solution for that could be to make the behavior configurable. 
Anyway, I wouldn't mind rolling my own `IObjectFactory` impl, but then I'd have to copy & paste the logic for resolving collection interfaces to implementations (which I also need in my application - again, for encapsulation). And duplicate code is usually a bad thing, even if it's semi-trivial. Maybe it would be a good idea to somehow factor out that logic, e.g. into an abstract base class?

EdwardCooke:
This feature was already implemented in a different, opt in way. This PR also appears to be abandoned, it's over nearly 7 years old. I'm going to close it.

PR #123, opened 2015-02-11
Fixed #121: Finding properties in parent interfaces

### LINKED ISSUE #121: Can't deserialize to interface that inherits properties from other interfaces
I would like to deserialize to an interface that inherits properties from other interfaces, but v3.5.1 does not currently appear to support that. Is there a possibility of adding support for this?

    public interface IBase
    {
        int Id { get; set; }
    }
    
    public interface IDerived : IBase
    {
        int SomeValue { get; set; }
    }
    
    public interface ITopLevel
    {
        IDerived SomeDerived { get; set; }
    }

    // Suppose that SomeObjectFactory maps interface types to concrete types
    var deserializer = new Deserializer(new SomeObjectFactory());
    
    // This triggers a SerializationException, "Property 'Id' not found on type 'IDerived'."
    var out = deserializer.Deserialize<ITopLevel>(someYamlWithId);

I believe this occurs because the `GetPublicProperties()` methods in [Portability.cs](https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Helpers/Portability.cs) do not detect when the provided type is an interface. If it's an interface, calling [Type.GetProperties()](https://msdn.microsoft.com/en-us/library/aky14axb) will only return the properties defined on the interface itself, and none of the inherited properties. To get properties an interface inherits from other interfaces you need to do something like:

    public static IEnumerable<PropertyInfo> GetPublicProperties(this Type type)
    {
        var bindingFlags = BindingFlags.Instance | BindingFlags.Public;
        if (!type.IsInterface) { return type.GetProperties(bindingFlags); }
        return (new Type[] { type })
            .Concat(type.GetInterfaces())
            .SelectMany(i => i.GetProperties(bindingFlags));
    }

(via [GetProperties() to return all properties for an interface inheritance hierarchy](http://stackoverflow.com/q/358835/111327) on StackOverflow)

emddudley:
I included a test that fails with the existing code but then succeeds with my changes. I modified the "portable" version of GetPublicProperties() as well, although I was not able to test it.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -320,12 +320,27 @@ namespace YamlDotNet.Test.Serialization
 		public Base BaseWithSerializeAs { get; set; }
 	}
 
-	public class Base
+	public class InterfaceExample
+	{
+		public IDerived Derived { get; set; }
+	}
+
+	public interface IBase
+	{
+		string BaseProperty { get; set; }
+	}
+
+	public interface IDerived : IBase
+	{
+		string DerivedProperty { get; set; }
+	}
+
+	public class Base : IBase
 	{
 		public string BaseProperty { get; set; }
 	}
 
-	public class Derived : Base
+	public class Derived : Base, IDerived
 	{
 		public string DerivedProperty { get; set; }
 	}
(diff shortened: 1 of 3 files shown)


aaubry:
This sounds good. I'm sorry that it took so long to review this PR. Thanks

Issue #122, opened 2015-02-10
New feature: Ability to specify scalar style in YAML output


matthewrwilton:
Hi

I'd like to be able to control what the scalar style of properties is in the serialised YAML. If you think such a feature is worthwhile I'd be happy to have a go at adding it in.

My thoughts on implementing this would be to add an additional property to the YamlMemberAttribute where the scalar style can be specified.

Is this something you'd be interested in receiving a pull request for?

aaubry:
Sure, this feature could be useful. I agree that YamlMember is the appropriate location to specify that. Thanks

matthewrwilton:
Great! I'll work on this and submit it as a pull request.

PR #125, opened 2015-02-24
Allow the style of scalar properties to be specified.

matthewrwilton:
Hi

As discussed in #122 these changes add a new property to the YamlMemberAttribute allowing the style of scalar values to specified.

This completely bypasses Emitter.SelectScalarStyle and does not check that the value is appropriate for the provided scalar style, leaving it up to the user to choose an appropriate style or ensure the the values are valid for the style. Is this suitable, or do you think we should check the values and scalar style to ensure  invalid YAML is not produced?

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -400,6 +400,22 @@ namespace YamlDotNet.Test.Serialization
 		}
 	}
 
+	public class ScalarStyleExample
+	{
+		public ScalarStyleExample()
+		{
+			var content = "Test";
+			this.LiteralString = content;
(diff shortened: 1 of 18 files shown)


aaubry:
I'm sorry I took so long to review this. I do not agree that this functionality should alter the interface of IObjectGraphVisitor. I have submitted another pull request, #128, which refactors this so that the style information flows inside IObjectDescriptor instead. Please take a look and let me know what you think. 
Thanks

matthewrwilton:
That's ok, I'm glad you've taken a look at it. Since it was my first change to the library I was a bit uncertain about where to locate some of the functionality, so I'll take a look at your changes.

wspresto:
It is 2020 and we still need this feature. What happened to this feature? What version still has this feature?

PR #127, opened 2015-03-28
Removing dependency to TypeInfo, now compiles in Unity w/ .net 2.0 subset

steinbitglis:
Hi! Thank you for sharing this super cool library!
I made it compile in Unity (Game Engine), even with some of the code compatibility options turned to "low". I'm doing this so people can use the library regardles of those settings. The only real issue that TypeInfo couldn't be found, and that was easy to work around. I don't know if you want this kind of feature, but I wanted to share my little discovery with the world :-)


aaubry:
You can't remove the dependency from TypeInfo from that code because it exists to build portable class libraries. That code is inside an `#ifdef PORTABLE` region that is active only when building the portable library. What are the requirements to compile in Unity?

steinbitglis:
I'm not sure exactly what all the requirements are, but the TypeInfo class doesn't exist there for sure. I replaced all the use of TypeInfo with similar calls to the Type class. Would this not be sufficient? I mean, I did compile with the PORTABLE flag after all. It's one of the reasons it was possible at all. Some functions seemed not to be used, so I just removed them. If they are necessary for other reasons... (tests or whatever) they should also be written in that way, if they also want to support Unity compilation in this way.

Another alternative is to put this code inside some new 'UNITY' preprocessor flag. Would that fit better?

aaubry:
The requirements are the following: The project has multiple build configurations, some of them target .NET 3.5, and other target the portable profile Profile259 (.NET Framework 4.5, Windows 8, Windows Phone 8.1, Windows Phone Silverlight 8).

If you specify the PORTABLE flag, you are saying that you want to build the portable version. On that profile, the regular reflection API is very limited, and one needs to use TypeInfo to achieve the same.

I know nothing about Unity, what are the requirements? Can't you use the YamlDotNet class library as-is? What are the constraints?

steinbitglis:
This is what I know from the top of my mind:
Unity uses a .net API that is very similar to .NET 3.5 in practice.

If I set the "Player Settings" to '.net 2.0' compatible (there are only two options, that one and '.net 2.0 subset'),  then YamlDotNet compiles fine without any changes. If I choose '.net 2.0 subset' then I get errors that refer to missing regular expression related stuff, until I set the 'PORTABLE' flag, which I found out about by reading your code, then the TypeInfo is the only problem.

From what I can read from microsofts own blog post, TypeInfo first came in .net 4.5, so Unity will never be able to compile that, unless they change their engine significantly.

aaubry:
Ok, I get that, but why are you specifying the `PORTABLE` flag? If you do omit it, the TypeInfo API won't be used and that should solve your problem.

steinbitglis:
Mostly out of suspicion.
In practice I could never get YamlDotNet to work on any aot compiled platform with Unity. It compiled, but crashed, so I'm looking for a solution (a 2 year old version of YamlDotNet btw.).

Also, I never (until now) got YamlDotNet to compile under the '.net 2.0 subset' setting, and I'm hoping that it's relevant to the crashes that I saw on the aot compiled platforms. It's also good to support this setting since I want to use YamlDotNet in a bug tracker that I'm building for Unity.

When I do try to compile with the ".net 2.0 subset", I seem to be missing regex-related and other classes. I don't know the full list of missing classes exactly, I just turned to the PORTABLE flag for an easy solution. I thought, "more portable... that sounds like what I need" :-)

After turning to many other JSON libraries for better Unity compatibility, I actually came back to YamlDotNet again. I know it works on the desktop platforms... and by the looks of it, now I can get it to work on the consoles as well, so I'm really happy. Crossed fingers and all. :-)

steinbitglis:
With this edit the YamlDotNet folder can be compiled within Unity if the 'UNITY' compile flag is defined. This must be done manually from within Unity.
I haven't changed YamlDotNet.Test or Readme.md, but I guess that might be something you might want to do as well.

aaubry:
I have reviewed your changes, and it seems to me that by attempting to reuse the code inside the "PORTABLE" sections, you end up duplicating lots of code in Portability.cs. I have made some changes to your pull request and committed those changes to [`pr/n127_steinbitglis`](https://github.com/aaubry/YamlDotNet/compare/pr/n127_steinbitglis) - 2eaa062bbe88212094d73d6da01a8b3379f17a9a. Can you checkout that code and validate that it works?

What I did was installing Install [Visual Studio 2013 Tools for Unity](https://visualstudiogallery.msdn.microsoft.com/20b80b8c-659b-45ef-96c1-437828fe7cf2), then:

* Create two new solution configurations:
  * `Debug-UnitySubset-v35`
  * and `Release-UnitySubset-v35`
* Make each of the above solution configurations targets `Unity Subset v3.5` and define the `UNITY` preprocessor constant.
  * This will allow anyone to easily build the project using Visual Studio, as well as to use continuous integration to build for the Unity target.
* Refactor the `#if ... #endif` blocks, mostly in `Portability.cs` to avoid duplicate code.

Please let me know if this works or if anything needs to be changed.

Thanks

PS: I wanted to update your PR with my changes, but I don't know how to do that, or if it is even possible, that's why I ended up committing a new branch.

steinbitglis:
Everything looks correct, I'll update the PR when I've tested it.
I hope I have time to test tomorrow, but no guarantees.

PR #128, opened 2015-03-29
Pr/n125 matthewrwilton

aaubry:
Replaces #125 with an alternative implementation that does not alter the contract of the object graph visitors.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -400,6 +400,22 @@ namespace YamlDotNet.Test.Serialization
 		}
 	}
 
+	public class ScalarStyleExample
+	{
+		public ScalarStyleExample()
+		{
+			var content = "Test";
+			this.LiteralString = content;
+			this.DoubleQuotedString = content;
+		}
+
+		[YamlMember(ScalarStyle = ScalarStyle.Literal)]
+		public String LiteralString { get; set; }
+
+		[YamlMember(ScalarStyle = ScalarStyle.DoubleQuoted)]
+		public String DoubleQuotedString { get; set; }
+	}
+
 	public class DefaultsExample
 	{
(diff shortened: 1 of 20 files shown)


matthewrwilton:
I was looking over the changes and I do have a question.

There was some code I had added in the SelectScalarStyle method in the Emitter class that would shortcut the scalar style selection if a style had been specified using the YamlMember attribute. That code has been removed and it is now possible for the scalar style to vary from what was specified.

In this case, should the property of the YamlMember attribute be renamed to something such as SuggestedScalarStyle or PreferedScalarStyle so that it is clearer that the final serialisation style may be different?


Also I am happy with you reverting the changes to the Emitter, so that it continues adjusting the scalar style. It was something I was uncertain about when implementing the feature originally, but I wanted to ensure that the scalar style the user specified was what the property was serialised as. I did realise this could result in invalid YAML being produced if the user was not careful and so can understand your decision to continue having the Emitter choose the final scalar style. Looking back I should have asked some questions to get a feel for how you want the library to behave.

Thanks

aaubry:
Regarding the changes to the Emitter, I am not sure about them. Maybe if the user has specified a style, the serializer should honor that, I don't know. However, to do that would only require to check if `Style != ScalarStyles.Any`; no additional property is needed.

matthewrwilton:
I think it is up to you and whether you think the library should always produce valid YAML, or if it is left up to users to ensure that the serialisation style is appropriate for the property and will produce valid YAML.

aaubry:
Let's leave it like this for now and we'll see if there is feedback when people try to use it.

matthewrwilton:
Sounds good to me. Thanks!

aaubry:
@wspresto, you're asking a question about somethig that was done 5 years ago. The branch was deleted because it was merged. Does this answer your question ?

wspresto:
Thank you for responding. No. I will be preparing an issue with detailed steps and examples on how to reproduce the bug.

PR #133, opened 2015-06-28
Fixed TypeConverter not finding implicit/explicit cast methods

ClockVapor:
TypeConverter was using Type.GetPublicMethods() which never finds op_Implicit or op_Explicit.

### DIFF
--- a/YamlDotNet/Serialization/Utilities/TypeConverter.cs
+++ b/YamlDotNet/Serialization/Utilities/TypeConverter.cs
@@ -187,7 +187,7 @@ namespace YamlDotNet.Serialization.Utilities
 			// Try to find a casting operator in the source or destination type
 			foreach (var type in new[] { sourceType, destinationType })
 			{
-				foreach (var method in type.GetPublicMethods())
+				foreach (var method in type.GetMethods(BindingFlags.Public | BindingFlags.Static))
 				{
 					var isCastingOperator =
 						method.IsSpecialName &&


aaubry:
Thanks for contributing. However, your change does not build on all targets, because the reflection API is different on the .NET Portable subset. This is the reason why there are extension methods that abstract the main code from these details. Also, there's no unit test to guarantee that the bug is fixed.
Can you fix the issues?

ClockVapor:
Sorry for the confusion. I see where this issue came from now. I'm using YamlDotNet for a Unity project, and it seems that the implementation of `GetPublicMethods()` is different in the Unity plugin on the Asset store. Here's what the plugin has:

```C#
public static IEnumerable<MethodInfo> GetPublicMethods(this Type type)
{
    if (type == null) throw new ArgumentNullException("type");
    return type.GetMethods(BindingFlags.Instance | BindingFlags.Public);
}
```
Whereas the lastest version up here has this:

```C#
#if PORTABLE
public static IEnumerable<MethodInfo> GetPublicMethods(this Type type)
{
    return type.GetRuntimeMethods()
        .Where(m => m.IsPublic && !m.IsStatic);
}
#else
public static IEnumerable<MethodInfo> GetPublicMethods(this Type type)
{
    return type.GetMethods(BindingFlags.Static | BindingFlags.Public);
}
#endif
```

The plugin is missing `BindingFlags.Static`, which is why my project failed to find `op_Implicit`/`op_Explicit`.

aaubry:
The Unity plugin is maintained by @steinbitglis. The code should be functionally equivalent, but there may be a few bugs here and there.

Actually, your analysis shows a difference between the PORTABLE code and the standard one: the standard code is requesting only static methods, whereas the portable one is excluding them. I will investigate this issue.

steinbitglis:
I'll check the asset store version when I get home from work :-)

steinbitglis:
I merged v3.6.1 from your repo into the Unity asset project, and it seems to be working fine in the editor.

Since I don't have a test pipeline for Unity builds, I will just run through the example's on a few AOT compiled builds tomorrow (those used to be the tricky ones).

The asset doesn't rely on Visual Studio, so there's a bit of cut'n paste here and there, just to remove the need for the 'UNITY' preprocessor flag. Also the 'AssemblyInfo.cs' seems to mess up for the compiler in Unity. Other than that, I'm trying to keep all the code identical.

For people that compile the library down to a managed dll, using Visual Studio, I guess they would be able to use your repo as-is. I don't know so much about that, since I don't use Visual Studio much.

steinbitglis:
Btw. Unity Asset Store usually takes a week to review the changes too :-p

aaubry:
FYI, the CI build also produces a [class library that targets Unity](https://ci.appveyor.com/project/aaubry/yamldotnet/build/3.6.1.117/artifacts). Since I have never worked with Unity, I am not sure if that is useful at all...

steinbitglis:
My guess is that it is.

Issue #113, opened 2014-12-11
Non-decimal integer bases are not supported


jeffman:
I have a simple class with one int property:

```csharp
class TestClass
{
    public int TestField { get; set; }
}
```

I try to deserialize the following YAML file:

```yaml
TestField: 0x10
````

...using the following code:

```csharp
var str = new StringReader("TestField: 0x10");
var d = new Deserializer();
var obj = d.Deserialize<TestClass>(str);
```

I would expect `obj.TestField` to have a value of 16. However, an exception is thrown during deserialization, when trying to parse `"0x10"` in `ScalarNodeDeserializer.cs`:

```csharp
case TypeCode.Int32:
	value = Int32.Parse(scalar.Value, numberFormat);
	break;
```

It seems that the possibility of having a non-decimal base is being ignored. Many integer bases (binary, octal, hexadecimal and sexagesimal) should be allowed per the [YAML spec](http://yaml.org/type/int.html). For reference, this works as expected in `pyyaml`:

```python
>>> import yaml
>>> print yaml.load("test: 0x10")
{'test': 16}
```

PR #139, opened 2015-09-14
Add support for different scalar integer bases

jeffman:
Addresses issue #113. Adds basic support for deserializing scalar integers
written in binary, octal, decimal, hex, and base 60, as allowed in the YAML
specification; see http://yaml.org/type/int.html. Adds unit tests for each
of these bases as well.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -63,6 +63,54 @@ namespace YamlDotNet.Test.Serialization
 		}
 
 		[Fact]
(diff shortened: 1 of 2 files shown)


jeffman:
Not sure why AppVeyor is failing; all tests pass on my local version and my changes aren't in the codepath for the failing test.

aaubry:
Maybe it is a bug in the test itself. I'll look into it.

aaubry:
I have noticed that some tests occasionally fail during continuous integration, but I have no idea why. I have updated the NuGet packages that are used in the test project and we'll see if that fixes it.

PR #141, opened 2015-09-17
Update TypeNameInTagNodeTypeResolver.cs

derFunk:
If the type could not be loaded, because it fails with an exception, make sure to return false to pass resolving to the next resolver in the list.


aaubry:
Nice one!

PR #160, opened 2016-01-22
Fixed Dictionary deserialization on AOT platforms (PS4 tested).

steinbitglis:
Hi!
We were having a lot of trouble with deserializing dictionaries on the PS4.
After a lot of digging, I found out that I couldn't call MethodInfo.MakeGenericMethod() ... unless the target method was cached from AOT compilation. As you probably know, JIT-compilation isn't allowed on AOT-based platforms.

As far as I know, there's no good reason why the same job can't be done with reflection in this case, since we're using tons of reflection anyway. This way, jit-compilation is avoided.

Feel free to comment, change and/or merge as you like.

I apologize for not adding unit tests and paying attention to formatting etc.
I felt it was important to share this improvement, since it fixes a real problem for many.


steinbitglis:
Uh... it seems that your tests don't like that I used Type.GetConstructor... but it's not a problem in Unity... what do I do?

aaubry:
This is due to Type.GetConstructor not being available on all platforms. Let me check that.

aaubry:
I'm trying to reproduce the problems that you encountered. I am compiling code using AOT, but I am never able to cause an error. I only got an AOT error once, but that was because I forgot to specify the "UNITY" define. Do you have an example or something that causes problems with AOT? Thanks

steinbitglis:
I encountered this during deserialization of a dictionary on the PS4 devkit.
I'll see if I can find something easier to replicate on monday. (Too busy this weekend, sorry).

steinbitglis:
I tried tons of stuff to compile c# mono code in aot-mode on windows, but it seems that mono doesn't support it on windows.

If you have a mac or linux machine.. you might be able to figure it out. If you want to try that, follow this link: http://www.mono-project.com/docs/advanced/aot/#full-aot

The best way I can come up with on windows is this:
1. Create a project in Unity
2. Install YamlDotNet for Unity from the Asset Store or use your own version and set the compile flag in Player Settings.
3. Add the following to test decryption of a dictionary (Deserializing_a_dictionary.cs):

```c#
using System.Collections.Generic;
using System.IO;
using System.Text;

using UnityEngine;

public class Deserializing_a_dictionary : MonoBehaviour {

    private readonly string doc =
@"myDictionary:
    winners: 3";

    void Start () {
        var input = new StringReader(doc);
        var deserializer = new Deserializer(namingConvention: new CamelCaseNamingConvention());
        var order = deserializer.Deserialize<MyObject>(input);

        Debug.Log("winners: " + order.myDictionary["winners"]);
    }
}

public class MyObject {
    public Dictionary<string, int> myDictionary { get; set; }
}
```

4. Put the `Deserializing_a_dictionary` script on any game object in the hierarchy.
5. Save the scene, include it in the build settings and compile for HTML5.
6. Open in Firefox Developer Edition
7. Press Ctrl-Shift-K to open the Web Console in Firefox, 
8. Something like this should appear: ExecutionEngineException: Attempting to call method 'YamlDotNet.Serialization.NodeDeserializers.GenericDictionaryNodeDeserializer::DeserializeHelper' for which no ahead of time (AOT) code was generated.

aaubry:
Thanks, this is exactly what I wanted. I have setup a build on a linux machine to consistently reproduce this issue, but for some reason I could not come up with code that triggered the error. I plan to use that as a test that guarantees that this problem is not reintroduced later.

aaubry:
I have setup a [build on Travis CI](https://travis-ci.org/aaubry/YamlDotNet) that identifies AOT incompatibility issues. What the build does is to compile a test program with mono and full AOT, then run it. The program runs a set of tests that cause the exceptions that your PR fixes. This should help preventing future regressions.

I have merged your code in the [branch](https://github.com/aaubry/YamlDotNet/tree/aot-tests) where I added the tests, named "[aot-tests](https://github.com/aaubry/YamlDotNet/tree/aot-tests)". I had to make a few changes because of incompatibilities with the "Portable" targets, and also because not every `IDictionary<,>` and `IList<>` implement `IDictionary` and `IList`, respectively.

If and when you have time, I would like if you could test that branch and confirm that it works without problems. Thanks

steinbitglis:
My dictionary deserialization test, seems to be running without any errors on the PS4, using  c3cc519e8c32ec36dccd928f054c05a5547a4a97.

aaubry:
That's good. I will merge this into master, then. If we encounter another incompatibility, we can add a test that triggers it and then fix it.

aaubry:
For reference, the merge commit is f3cbdc93b7a041336eb5ce827d7b8409bbed8606.

steinbitglis:
Great work! Is this a good time to update the Asset Store?

aaubry:
Yes, I think so. This version shoud be more useful for Unity developers.

steinbitglis:
.. and it has been accepted :-) I will be following your version numbers from now on, to make it more apparent to Unity users what's going on behind the asset.

PR #164, opened 2016-02-04
Modify YamlVisitor behavior

nbarbettini:
I made a change to YamlVisitor to make it behave more like the ExpressionVisitor class: by overriding the `Visit` methods, a subclass can change the visitor's behavior. (For example, choosing not to visit a particular mapping.)

I also broke out visiting the key and value of a mapping into the `VisitPair` method to make it easier to observe this behavior.

These changes were super helpful when I was writing a custom YAML configuration parser in my project. (The result is [here](https://github.com/stormpath/stormpath-dotnet-config/blob/6c8ce545bdcac11f8138611557e2d36628917321/Microsoft.Extensions.Configuration.Contrib.Stormpath.Yaml/Visitor/ContextAwareVisitor.cs).) If you don't want to introduce these changes into your code, no worries - thought I'd share in case it was helpful to someone else. :smile:

### DIFF
--- a/YamlDotNet/RepresentationModel/YamlVisitor.cs
+++ b/YamlDotNet/RepresentationModel/YamlVisitor.cs
@@ -1,16 +1,16 @@
 //  This file is part of YamlDotNet - A .NET library for YAML.
 //  Copyright (c) Antoine Aubry and contributors
-    
+
 //  Permission is hereby granted, free of charge, to any person obtaining a copy of
 //  this software and associated documentation files (the "Software"), to deal in
 //  the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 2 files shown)


aaubry:
This looks good. The only problem is that it is a breaking change. Maybe we should introduce this as a new class, and mark the old one as `[Obsolete]`, so we don't break existing implementations. Does this make sense?

nbarbettini:
Yeah, I agree. In my code, I named this implementation `YamlVisitorBase` - shall I do that here (and mark the other obsolete)?

aaubry:
Sure, that name sounds good. Please do that. Thanks

nbarbettini:
Done :+1:

PR #169, opened 2016-03-07
Why is that null char here in the first place ?

gissehel:
Here is my pull request to try to fix the problem exposed in http://stackoverflow.com/questions/35681811/null-char-in-output-yaml-in-yamldotnet-serialization

(Unexpected `'\0'` char in yaml output when a scalar starts with a new line)

Here is a `dotnetfiddle.net` that shows the problem: https://dotnetfiddle.net/jqVcL3 

I don't understand the presence of the \0 in the source, so I assume it was a typo.

### DIFF
--- a/YamlDotNet/Core/Emitter.cs
+++ b/YamlDotNet/Core/Emitter.cs
@@ -1702,7 +1702,7 @@ namespace YamlDotNet.Core
 
             if (analyzer.IsSpace() || analyzer.IsBreak())
             {
-                var indentHint = string.Format(CultureInfo.InvariantCulture, "{0}\0", bestIndent);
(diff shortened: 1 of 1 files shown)


aaubry:
Thanks for the fix. I checked the history of the file, but I could not find any reason why that code would justify that null char. The weird thing is that this code is very old and still did not cause trouble until now.

CrustyJew:
@aaubry this bug seems to still exist as I'm getting `'\0'` char emitted still. And I see the 

` var indentHint = string.Format(CultureInfo.InvariantCulture, "{0}\0", bestIndent);`

still exists under `WriteBlockScalarHints(string value)`

Can we get this merged in?

PR #173, opened 2016-04-14
Ignore enum value case during deserialization

roman-yagodin:
The enum value case is ignored by TypeConverter class (from Serialization/Utilities), but not by ScalarNodeDeserializer.


aaubry:
This makes sense. Thanks.

PR #174, opened 2016-04-14
Yaml Attribute overrides

dol-leodagan:
Implementation of Serializer Attribute Overriding Feature.

I wanted to switch from .NET XML Serializer to YamlDotNet Serializer but there is a lack of Seralizer customization...

I was using intensively XML Attribute Overrides with the XML Serializer to map classes that didn't implement specific attributes, this is useful to serialize class by reflection and have format agnostic importer/exporter !

In this PR I tried to rely as most as possible on existing inner behavior of YamlDotNet to rely on existing ```YamlAttributesTypeInspector``` and provide overriding for any current or future Yaml Attribute.

I provided a few Unit Tests.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -251,6 +251,30 @@ namespace YamlDotNet.Test.Serialization
 
             output.AliasTest.Should().Be(input.AliasTest);
         }
+        
+        [Fact]
+        public void RoundtripAliasOverride()
+        {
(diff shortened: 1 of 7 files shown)


aaubry:
I only found time to review your changes now, sorry about that.
Everything looks good, so I will accept the PR. Thanks for making the effort to understand the internal type descriptor mechanism and staying aligned with the architecture.

dol-leodagan:
No problem for the delay.

It's good to know the efforts I made to match the code structure were useful to merge this feature easily !

PR #175, opened 2016-04-18
Add a YamlNodeType enumeration property to nodes

kzu:
This makes processing easier than just trying to use as/is
all over when processing a doc. This is similar to the way
System.Linq.Expressions namespace does it.


aaubry:
Thanks, I think this will be useful.

Issue #177, opened 2016-05-05
Serializing double.MaxValue is incorrect


vwxyzh:
double.MaxValue will be serialized as `1.79769313486232E+308`
but the real value should be `1.7976931348623157E+308`, when deserializing, it will throw OverflowException.

repro:
```c#
var s = new Serializer();
var sw = new StringWriter();
s.Serialize(sw, new Dictionary<string, double> { ["a"] = double.MaxValue });
Console.WriteLine(sw.ToString());
var d = new Deserializer();
d.Deserialize<Dictionary<string, double>>(new StringReader(sw.ToString()));
```

PR #178, opened 2016-05-06
fix serializing double.MaxValue #177

vwxyzh:
use "G17", see: [Standard Numeric Format Strings](https://msdn.microsoft.com/en-us/library/dwhawy9k(v=vs.110).aspx)

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -1016,25 +1016,45 @@ namespace YamlDotNet.Test.Serialization
         }
 
         [Fact]
-        public void SpecialFloatsAreSerializedCorrectly()
+        public void SpecialDoublesAreSerializedAndDeserializedCorrectly()
         {
-            var serializer = new Serializer();
-
-            var buffer = new StringWriter();
-            serializer.Serialize(buffer, new double[]
+            var expected = new double[]
             {
-                double.NaN,
-                double.PositiveInfinity,
-                double.NegativeInfinity,
-            });
+                double.MaxValue,
+                double.MinValue
+            };
+
+            var serializer = new Serializer();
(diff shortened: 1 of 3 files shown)


aaubry:
Hi! Thanks for contributing. Could you add unit test(s) for this feature? This is necessary to ensure that this won't be broken on a future change.

vwxyzh:
unit test added.

aaubry:
I have another question. Shouldn't we use the "R" (roundtrip) format spectfier instead?

> The round-trip specifier guarantees that a numeric value converted to a string will be parsed back into the same numeric value. When a numeric value is formatted using this specifier, it is first tested using the general format, with 15 spaces of precision for a Double and 7 spaces of precision for a Single. If the value is successfully parsed back to the same numeric value, it is formatted using the general format specifier. However, if the value is not successfully parsed back to the same numeric value, then the value is formatted using 17 digits of precision for a Double and 9 digits of precision for a Single. Although a precision specifier can be appended to the round-trip format specifier, it is ignored. Round trips are given precedence over precision when using this specifier. This format is supported by floating-point types only.

http://www.hanselman.com/blog/WhyYouCantDoubleParseDoubleMaxValueToStringOrSystemOverloadExceptionsWhenUsingDoubleParse.aspx

vwxyzh:
in [Standard Numeric Format Strings](https://msdn.microsoft.com/en-us/library/dwhawy9k(v=vs.110).aspx)
>**Important**
>In some cases, Double values formatted with the "R" standard numeric format string do not successfully round-trip if compiled using the /platform:x64 or /platform:anycpu switches and run on 64-bit systems. See the following paragraph for more information.

aaubry:
Thanks for the clarification. I had to refactor the tests, because your changes deleted important tests that checked the correct serialization of constants such as infinitty.

Issue #135, opened 2015-07-06
Serialization Error when string start with quote (')


vicancy:
Quote should be escaped.
`SyntaxErrorException` throws when I serialize and deserialize a string `'a`.

**REPRO**
```csharp
using System.IO;
using YamlDotNet.Serialization;

public class Program
    {
        static void Main(string[] args)
        {
            string k = "'a";
            using (StringWriter writer = new StringWriter())
            {
                new Serializer().Serialize(writer, k);
                using (StringReader reader = new StringReader(writer.ToString()))
                {
                    var j = new Deserializer().Deserialize<string>(reader);
                }
            }
        }
    }
```

PR #184, opened 2016-06-13
Issue 135 single quote serialization error

Archanian:
Fix for issue #135 whereby strings containing single quotes are serialized incorrectly.


Archanian:
My solution here is to double quote any strings containing single quotes. There may be a case for escaping single quotes (ie. ''), but I feel this is a more robust solution.

aaubry:
Hi, thanks for contributing.
I agree that this is a bug that needs fixing, but if you fix that at the serialization level, any code that uses the emitter will still have the bug. 

The correct place to fix this is in the Emitter class, which takes care of checking the content of the scalar to help select its style. I think that the AnalyzeScalar needs to be changed [here](https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Core/Emitter.cs#L293).

Would you like to attempt to fix the bug there?

Archanian:
Sure thing, I can have a go at fixing it there, although I am a little unsure about that method and all the flags that are set ... I am assuming I should do the same string check and set scalarData.style = ScalarStyle.DoubleQuoted?

Cheers.

aaubry:
I think it should be enough to add the single quote character to the list of characters that are checked [here](https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Core/Emitter.cs#L343). This sets flags that influence the selection of the scalar style in the SelectScalarStyle method.

I'm sorry about the messy code. This part is a translation of the libyaml C code to C#, and at the time I did not understand it well, and I wanted to keep it as close as possible to the original code so that it would be easy to keep it up-to-date.

Archanian:
OK, adding single quotes to that list of characters results in the style being set to ScalarStyle.SingleQuoted ... because it hits this check:
`if ((flowLevel != 0 && !scalarData.isFlowPlainAllowed) || (flowLevel == 0 && !scalarData.isBlockPlainAllowed))
                {
                    style = ScalarStyle.SingleQuoted;
                }`

Archanian:
Let me know what you think of the above changes

aaubry:
You can't set `isSingleQuotedAllowed` to false, because that would break code that explicitly selects that style. Why not let it use single quoted and let the code escape the quotes?

Archanian:
OK, but if a string contains single quotes, should single quotes be allowed anyway? I'm not quite sure what you're suggesting, but I'm fairly certain just adding single quote to the buffer check does not work.

aaubry:
If the string contains single quotes, they will be escaped by the emitter. E.g:
`hello 'world'`
will become
`'hello ''world'''`
Of course, in this specific case it would be prettier to use double quotes for escaping. One possibility would be to change the style selection in [SelectScalarStyle](https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Core/Emitter.cs#L874) to choose either single or double quoted depending on the content of the scalar, but only when the requested style was "Any". What do you think?

Archanian:
OK that sounds good, I've updated the pull request, setting the style where you suggested. I'm not sure what combination of factors sets the requested style to "Any", do you have more information you can give me on that?

Archanian:
Side note: I could not use the test helper to create a Scalar with styles "Any", as it resulted in noTag = true, isPlainImplicit = false, isQuotedImplicit = false, which causes the Emitter to throw an exception (line 871).

aaubry:
The scalar style "Any" is the default one, so if you don't specify anything, that will be the selected style. Basically, it says "pick the best style for me". I hope to review your changes later today.

aaubry:
I made a small change to your code to keep the previous behavior for multiline scalars. Would you mind to take a look to #187 ?

Archanian:
Cool that looks good, thanks for your help.

aaubry:
I have merged the changes. Thanks for your help!

aaubry:
I plan to release a new version at the beginning of July, and this will be included in it.

PR #187, opened 2016-06-24
Archanian issue 135 single quote serialization error

aaubry:
Improve PR #184


Archanian:
Looks good to me, thanks.

aaubry:
The build did fail, but it was not related with the code, so I will ignore the error.

PR #193, opened 2016-07-08
Fix issue #190

### LINKED ISSUE #190: OverridePropertyDescriptor doesn't lookup properties properly.
Currently, the `Deserializer` creates a type inspector like this.

    typeDescriptor.TypeDescriptor =
        new CachedTypeInspector(
            new YamlAttributesTypeInspector(
                new YamlAttributeOverridesInspector(
                    new NamingConventionTypeInspector(
                        new ReadableAndWritablePropertiesTypeInspector(
                            new ReadablePropertiesTypeInspector(
                                new StaticTypeResolver()
                            )
                        ),
                        namingConvention
                    ),
                    overrides
                )
            )
        );

Using a simple class like this:
```
public class Foo
{
	public bool IsRequired { get; set; }
}
```

And overrides:

```
var yao = new YamlAttributeOverrides();
var isRequiredAttribute = new YamlMemberAttribute();
isRequiredAttribute.Alias = "Required";
yao.Add(typeof(Foo), "IsRequired", isRequiredAttribute);
```

The `NamingConventionTypeInspector` returns an `IPropertyDescriptor` with the name **isRequired**.  Then the `YamlAttributeOverridesInspector` wraps this with an `OverridePropertyDescriptor`.  Later on in the serialization process `GetCustomAttribute` is called on the descriptor which checks the types (`Foo` in this case) for properties with the name `isRequired`.  This returns nothing because of the casing mismatch and as a result it skips the AttributeOverride lookup.  

So, that's the whole issue and I believe the solution is relatively simple.  Swapping the order of the `YamlAttributeOverridesInspector` and `NamingConventionTypeInspector` should be enough to resolve the issue.  I would assert that the `NamingConventionTypeInspector` should be placed after **both** the `YamlAttributesTypeInspector` and `YamlAttributeOverridesInspector` so that naming conventions are applied to overrides as well.

The `GetProperties` calls on the inspectors give this chain:

**YamlAttributeOverridesInspector (and lower):** IsRequired 
**YamlAttributesTypeInspector:** Required
**NamingConventionTypeInspector:** required 

Thoughts?

aaubry:
This fixes #190

### DIFF
--- a/YamlDotNet/Serialization/Deserializer.cs
+++ b/YamlDotNet/Serialization/Deserializer.cs
@@ -90,18 +90,18 @@ namespace YamlDotNet.Serialization
 
                 typeDescriptor.TypeDescriptor =
                     new CachedTypeInspector(
-                        new YamlAttributesTypeInspector(
-                            new YamlAttributeOverridesInspector(
-                                new NamingConventionTypeInspector(
+                        new NamingConventionTypeInspector(
+                            new YamlAttributesTypeInspector(
+                                new YamlAttributeOverridesInspector(
                                     new ReadableAndWritablePropertiesTypeInspector(
                                         new ReadablePropertiesTypeInspector(
                                             new StaticTypeResolver()
                                         )
                                     ),
-                                    namingConvention
-                                ),
-                                overrides
-                            )
+                                    overrides
+                                )
+                            ),
+                            namingConvention
                         )
                     );
 
@@ -244,6 +244,14 @@ namespace YamlDotNet.Serialization
             return new Deserializer(valueDeserializer);
         }
 
+        public T Deserialize<T>(string input)
+        {
+            using (var reader = new StringReader(input))
+            {
+                return (T)Deserialize(reader, typeof(T));
+            }
+        }
+
         public T Deserialize<T>(TextReader input)
         {
             return (T)Deserialize(input, typeof(T));
@@ -252,6 +260,14 @@ namespace YamlDotNet.Serialization
         public object Deserialize(TextReader input)
         {
             return Deserialize(input, typeof(object));
+        }
+
+        public object Deserialize(string input, Type type)
+        {
+            using (var reader = new StringReader(input))
+            {
+                return Deserialize(reader, type);
+            }
         }
 
         public object Deserialize(TextReader input, Type type)
--- a/YamlDotNet/Serialization/YamlAttributeOverrides.cs
+++ b/YamlDotNet/Serialization/YamlAttributeOverrides.cs
@@ -93,14 +93,14 @@ namespace YamlDotNet.Serialization
                 while (currentType != null)
                 {
                     ++currentPriority;
-                    if (currentType == matchType)
+                    if (currentType == RegisteredType)
                     {
                         return currentPriority;
                     }
                     currentType = currentType.BaseType();
                 }
 
-                if (matchType.GetInterfaces().Contains(matchType))
+                if (matchType.GetInterfaces().Contains(RegisteredType))
                 {
                     return currentPriority;
                 }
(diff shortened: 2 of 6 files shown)


YamlDotNet/Serialization/Deserializer.cs
```diff
@@ -85,16 +85,16 @@ public IPropertyDescriptor GetProperty(Type type, object container, string name,
             typeDescriptor.TypeDescriptor =
                 new CachedTypeInspector(
                     new YamlAttributesTypeInspector(
-                        new YamlAttributeOverridesInspector(
-                            new NamingConventionTypeInspector(
+                        new NamingConventionTypeInspector(
```
veleek: Do you want to move NamingConventionTypeInspector outside of YamlAttributesTypeInspector as well?  With it inside the behavior is inconsistent.

If I have a property name FooBar with an attribute alias Foobar, and I use 'LowerCaseNamingConvention' then I believe the resulting output will be Foobar still (because it's applied after the naming convention).  But if you provide an AttributeOverride with the samething, the output will be foobar because the NamingConvention is applied after the attribute overrides.  I haven't validated this yet, but you can probably verify faster than I can.

YamlDotNet/Serialization/YamlAttributeOverridesInspector.cs
```diff
@@ -93,18 +93,9 @@ public void Write(object target, object value)
...
-                if (prop != null)
-                {
-                    var attr = overrides.GetAttribute<T>(prop.DeclaringType, Name);
-                    if (attr != null)
-                        return attr;
-                }
-                
-                // Default to base behavior
-                return baseDescriptor.GetCustomAttribute<T>();
+            {
```
veleek: Ahhh, MUCH cleaner. :)

YamlDotNet/Serialization/Deserializer.cs
aaubry: I was also concerned about this, but this order works because the overrides does not change the name of the property, it adds a new attribute to the list, and then YamlAttributesTypeInspector will use the name from the attribute to override the naming convention.

YamlDotNet/Serialization/YamlAttributeOverridesInspector.cs
aaubry: Yes, I don't know why that code was reading the property again. It did not make sense to me.

YamlDotNet/Serialization/YamlAttributeOverrides.cs
```diff
@@ -1,93 +1,182 @@
...
+            /// </summary>
+            /// <returns>The priority of the match. Higher values have more priority. Zero indicates no match.</returns>
+            public int Matches(Type matchType)
+            {
+                var currentPriority = 0;
+                var currentType = matchType;
+                while (currentType != null)
+                {
+                    ++currentPriority;
+                    if (currentType == matchType)
```
veleek: You're setting currentType to matchType at the top of this method so this matches immediately and caused it to return 1.

YamlDotNet/Serialization/YamlAttributeOverrides.cs
```diff
@@ -1,93 +1,182 @@
...
+                    {
+                        return currentPriority;
+                    }
+                    currentType = currentType.BaseType;
+                }
+
+                if (matchType.GetInterfaces().Contains(matchType))
```
veleek: Is this what we want to do?  Should one of these be ReferenceType?

`ReferenceType.GetInterfaces().Contains(matchType)`?

YamlDotNet/Serialization/YamlAttributeOverrides.cs
aaubry: Fixed. Thanks.

YamlDotNet/Serialization/YamlAttributeOverrides.cs
aaubry: Fixed. Thanks.

PR #203, opened 2016-08-25
Attempt to fix EventReader limitations.

aaubry:


joaope:
Brilliant! Less wrapping is always good. And the breaking change is less visible as you use IParser all the way through.

joaope:
How/where do you usually document this breaking changes? If you do it at all.

Let me know if I can help to get this merged asap. Tks.

aaubry:
The breaking changes are documented in the release notes, and also expressed in the version number. In this case, the major version number will be incremented, signaling a breaking change.
I plan to merge this later today unless I get feedback from other people, and I plan to release the next version, which will include this change, during this week. I have some other changes that I would like to include, and I need some more work to merge them.

PR #204, opened 2016-08-31
Serializer and deserializer builders

aaubry:
This is an important change that adds "builders" that can be used to configure the Serializer and Deserializer through a fluent syntax. The main objective of this is to allow more control over the composition of services performed by these two classes. This means that every aspect of the composition should be extensible / overridable. This means that things like injecting a custom TypeInspector or replacing the the default ArrayNodeDeserializer with an alternative implementation become possible and easy.

In order to avoid breaking existing code, the constructors of Serializer and Deserializer have been kept but marked as obsolete. In a future release they will be discarded, but for now existing code should not need to be changed.


tomitrescak:
Hi, thanks for the hard work! Was wondering if this is documented somewhere? All the docs I've found are using the obsolete API

Issue #189, opened 2016-06-28
Serialising a string starting with a whitespace and containing a new line results in malformed YAML


neuroid:
Hi,

Consider the following example:

```csharp
using System;
using YamlDotNet.Serialization;

namespace Test
{
    internal class Program
    {
        private static void Main(string[] args)
        {
            var serializer = new Serializer();
            serializer.Serialize(Console.Out, " line\nline");
        }
    }
}
```

The above results in the following output:

```
>2-
   line
  line
```

There is actually a `NUL` byte after `>2`:

```
00000000  3e 32 00 2d 0d 0a 20 20  20 6c 69 6e 65 0d 0a 20  |>2.-..   line.. |
00000010  20 6c 69 6e 65 0d 0a                              | line..|
00000017
```

Tested with YamlDotNet 3.9.0 from NuGet.

neuroid:
@aaubry Any chance of releasing a 4.0.1 to NuGet, containing this fix? :)

neuroid:
I'm closing since the issue is now fixed and prereleases are available on NuGet.

aaubry:
Hi,
I will do a release in the next few days that will include this fix. Thanks.

PR #205, opened 2016-09-09
Fix Issue 189 extra '\0' after indentation indicators

asmichi:
Hi!

This fixes issue #189 by removing extra '\0' after indentation indicators. The extra '\0' which is not allowed here makes resulting YAML text malformed.

Could you take a look? Thanks!


aaubry:
This looks good. Thanks!

Issue #180, opened 2016-05-23
Support .NET Core


nbarbettini:
It'd be great to have .NET Core support for YamlDotNet now that RC2 is here.

aaubry:
Unless I am mistaken, there is already Support for .Net core. The NuGet package contains the necessary targets, and I successfully compiled an app using a previous release. 
Please validate and if .Net core support does not work, provide more details of what you tried and which problems you encountered. Thanks

nbarbettini:
Trying to add YamlDotNet 3.8.0 to a .NET Core RC2 library project results in this error:

```
    Package YamlDotNet 3.8.0 is not compatible with netstandard1.3 (.NETStandard,Version=v1.3). Package YamlDotNet 3.8.0 supports:
      - dotnet (.NETPlatform,Version=v5.0)
      - net35 (.NETFramework,Version=v3.5)
      - portable-net45+win8+wp8+wpa81 (.NETPortable,Version=v0.0,Profile=Profile259)
    One or more packages are incompatible with .NETStandard,Version=v1.3.
```

The project would need to target some `netstandardX.X` in order to be installed in .NET Core projects. See here: http://dotnet.github.io/docs/core-concepts/libraries/libraries-with-cli.html#how-to-target-net-core

nbarbettini:
For now, it's possible to work around by adding `imports: [ "dotnet" ]` to my project.json 👍

adamhathcock:
Made a PR for RTM.  Would like to get rid of the import statement.

sparksterz:
@nbarbettini I'm now trying to do this with .net core 1.1 but I'm unable to simply add `imports: [ "dotnet" ]` anywhere and have it succeed...is there something I'm missing?

aensidhe:
@aaubry I've made a [dirty port](https://github.com/aaubry/YamlDotNet/compare/master...e-vote:master) of library and tests to preview4 tooling of MSBuild tools. If you need it, feel free to use it.

- Perfomance and sample are not ported
- Unity and portable not checked
- System.Drawing tests excluded from netcore tests

We plan to use it until official release of netcore support.

Nothing4You:
Since #244 is done I think this can be closed

aaubry:
[...]

PR #210, opened 2016-10-06
Create NetCore project for .NET Core RTM

adamhathcock:
In the vein of: https://github.com/aaubry/YamlDotNet/issues/180 and https://github.com/aaubry/YamlDotNet/pull/181

I made a separate project and directory for a net core build.  Though I'd recommend consolidating everything into one xproj/project.json.  I didn't want to do it all at once as it seems your build has quite a few things that need to get put into project.json.  You can use mine as a starting point to put even `net20` and `net35` targets.

All your nuspec and options in csprojs can go into a project.json.  Here's the spec: https://docs.microsoft.com/en-us/dotnet/articles/core/tools/project-json

You can update your tooling as well here: https://www.microsoft.com/net

You're correct in that your code supports .NET Core.  However, the nuget targetting is off in that you have use an `import` statement for `dotnet` to do so.  You might could just get away with changing the `dotnet` in your nuspec files to `netstandard1.0` but it would be questionable if that works 100%.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ README.html
 
 samples/dotnet/project.lock.json
 YamlDotNet/Properties/AssemblyInfo.Generated.cs
+project.lock.json
--- a/YamlDotNet.NetCore/YamlDotNet.NetCore.xproj
+++ b/YamlDotNet.NetCore/YamlDotNet.NetCore.xproj
@@ -1,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
+
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>20c5d482-25ae-4096-a4c1-42a678348cf7</ProjectGuid>
+    <RootNamespace>YamlDotNet.NetCore</RootNamespace>
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">.\obj</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">.\bin\</OutputPath>
+    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <SchemaVersion>2.0</SchemaVersion>
+  </PropertyGroup>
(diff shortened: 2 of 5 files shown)


adamhathcock:
The idea as to jump start changing the build configurations into a more cross platform way.  Ideally you'd use xproj/project.json everywhere to do what you're doing now.  Then when the new tooling is release it will convert to the new csproj from format.

I was attempting to do a minimal change.  I didn't look at the solution file and notice it adding those changes.  I apologize for how messy it is.  I'll try to clean it up soon.

The reason for doing this is that I don't believe that simply copying the portable build you have and marking it as netstandard is good enough and not correct.  Incorporating that target into a cross platform build would help me and others.

aaubry:
When I added the "dotnet" target, I followed the guidance from https://oren.codes/2015/07/29/targeting-net-core/
According to that post, the existing portable library should work out-of-the-box. It is just a matter of placing them in `lib\dotnet` and adding the correct dependencies to the nuspec file. I did that and then created a .NET core project that used the package without issues, so I don't see what more needs to be done, apart from changing to `netstandard1.0`.

Regarding the xproj/project.json, what are the benefits of having two different projects to build the same code? The existing csproj is configured to produce builds for all the supported platforms, signed and not signed, and it compiles on windows, mono, and on the two CI services. What value do the new project files add to this?

adamhathcock:
That post is over a year old and .NET Core has evolved much since then.  Though it's stable as of a few months ago.  They're going to introduce a .NET Standard 2.0 which may change things again.  Having a real build versus hacking in something that happens to work now seems like the correct thing to do.

xproj works for all platform csproj did as well as the new .NET Core platforms.  That's the additional value.  If you just install the tooling from http://dot.net it works well with Visual Studio 2015.

aaubry:
I'm sorry, I was unable to answer sooner. At the moment, I am unable to invest time to make this changes. However, I don't mind if, as part of this pull request, the projects are migrated to xproj / project.json, as long as the following stays true:

- One **single** project to build to every configuration:
  - **Debug** - for development
  - **Debug-AOT** - used to run tests that ensure compatibility with mono's AOT compilation
  - **Debug-UnitySubset-v35** - debug build for Unity target
  - **PerformanceTests** - release build used to run the performance tests
  - **Release-Signed** - release build that targets .NET framework 3.5, signed (we could consider upgrading to .NET 4.5)
  - **Release-Unsigned** - release build that targets .NET framework 3.5, not signed (we could consider upgrading to .NET 4.5)
  - **Release-Portable-Signed** - release build that targets .NET framework Profile259, signed
  - **Release-Portable-Unsigned** - release build that targets .NET framework Profile259, not signed
  - **Release-UnitySubset-v35** - release build for Unity target
- Ability to open, build and debug the solution on windows, either with a free edition of Visual Studio 2015 or with Visual Studio Code.
- Ability to open, build and debug the solution on linux, either with Monodevelop or with Visual Studio Code.
- The CI on AppVeyor must pass. This means that:
  - it is able to build the following targets: Release-Unsigned, Release-Signed, Release-Portable-Unsigned, Release-Portable-Signed and Release-UnitySubset-v35.
  - it produces two NuGet packages, one containing the signed versions, the other containing the unsigned versions.
- The CI build on Travis must pass. This means that it is able to compile the solution using mono with `--aot=full`, then run the test with `--full-aot`.

If you, or someone else, are willing to make these changes and make sure that everything in this list stays true, I will gladly accept the changes. I just don't have time to make them myself.

aaubry:
I am closing this as .NET Core support of this was completed on [another PR](https://github.com/aaubry/YamlDotNet/pull/255), which has been merged into master. I have released version [4.2.0](https://www.nuget.org/packages/YamlDotNet/4.2.0) which contains the new `netstandard1.3` assembly.
If you find anything that is not as it should, please open an issue.

PR #222, opened 2016-11-28
Allow anchors redefinition/overwriting

mlaily:
This pull request allows anchors to be redefined further down the stream instead of throwing a `DuplicateAnchorException`.

From [the 1.1 spec](http://yaml.org/spec/1.1/#id863390) as well as [the 1.2 spec](http://www.yaml.org/spec/1.2/spec.html#id2765878):
> #### 3.2.2.2. Anchors and Aliases
> When composing a representation graph from serialized events, an alias node refers to the most recent node in the serialization having the specified anchor. Therefore, anchors need not be unique within a serialization.

This pull request also generates more human-friendly anchors (of the form "A", "B", ..., "AA", "AB"...) when saving a stream - see the last commit.

Don't be too afraid by the added lines count. Most of it is from yaml test data :)


aaubry:
This looks good, except for the anchor names part. The name generation algorithm seems more complex than it needs to be, and I am not sure that generating anchors in caps is more human-friendly than numbers.
Maybe I could merge the anchor redefinition code and we can discuss the anchor names separately?

mlaily:
Fair enough. I removed the last commit.

The "human-friendly" side is more about generating sequential combinations of letters rather than random numbers. 
And I agree the algorithm looks complex, but I'm not sure it's avoidable if we also want it to be efficient.

PR #223, opened 2016-11-29
Avoid crashing with a StackOverflowException when iterating over the AllNodes property when it's infinitely recursive

mlaily:
The implementations of `AllNodes` are delegated to an `abstract internal IEnumerable<YamlNode> SafeAllNodes(RecursionLevel level)` method that keeps track of the current recursion level and throws a `MaximumRecursionLevelReachedException` if the maximum recursion level is reached.

The maximum recursion level is currently set to 1000.

### DIFF
--- a/YamlDotNet.Test/RepresentationModel/YamlStreamTests.cs
+++ b/YamlDotNet.Test/RepresentationModel/YamlStreamTests.cs
@@ -52,13 +52,24 @@ namespace YamlDotNet.Test.RepresentationModel
             // of just on AllNodes because in debug mode, AllNodes
             // is accessed after the document loads to check for unresolved aliases...
             var loadAndAccessAllNodes = new Action(() =>
-                {
-                    var stream = new YamlStream();
-                    stream.Load(Yaml.ParserForText("&a [*a]"));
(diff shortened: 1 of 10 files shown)


mlaily:
I also implemented the same strategy in the `YamlNode.ToString()` methods, but instead of throwing an exception when the node is infinitely recursive, it breaks the infinite recursion by returning an error message as a value. (See the last three commits)

aaubry:
I'm not sure about this. Maybe `AllNodes` should instead return each node only once?
What is the usefulness of this property if there is no guarantee that it will work for a given document?

mlaily:
I don't know about the usefulness of `AllNodes`, but I'm pretty sure it's not supposed to crash the application if there is some infinite recursion :)

Returning each node only once would mean keeping a list of already returned nodes? This seems like a bad idea.

At least, with the code in this pull request, there is a catchable exception.

aaubry:
Sorry, I did not have time to review this earlier. I agree that it is better if AllNodes and ToString do not crash. I'll leave the discussion about the usefulness of AllNodes for later. Thanks

mlaily:
Thank you!

PR #229, opened 2016-12-17
WithTagMapping for Serializer

trzepka:
Resolves issue 145.

Adds a new event emitter that on serialisation will set the right tag for registered types. Use SerializerBuilder.WithTagMapping() to register a new tag.


aaubry:
Nice one. Thanks!

Issue #231, opened 2017-01-08
DateTimeConverter


justinyoo:
Hi,

I know there's only one converter out-of-the-box, ie. [GuidConverter](https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Serialization/Converters/GuidConverter.cs).

YamlDotNet itself can convert string to DateTime automatically and vice-versa. However, if the date/time format is not that expected, it fails conversion. Would it be part of your roadmap to include those frequently used converters within the library at some stage? I can send a PR for `DateTimeConverter`, if you like.

It's probably related to #196 (DateTimeOffset) issue by the way.

Cheers,

aaubry:
Hi,

Sure, it would be good to have a DateTimeConverter so that we can configure the conversion behavior. Please send a PR that adds it. Thanks

justinyoo:
@aaubry Sure. Will send PR soon. Thanks for accepting my suggestion.

PR #234, opened 2017-01-11
Add DateTimeConverter

justinyoo:
As discussed in #231 , I'd like to send this PR for `DateTimeConverter`. It would be great if you have a review and merge it.

What it does are:

* It accepts [`DateTimeKind.Utc`](https://msdn.microsoft.com/en-us/library/shx7s921(v=vs.110).aspx) and [Standard Date and Time Format Strings](https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx) of "G" as its default parameters, if they are omitted.
* For deserialisation, it accepts as many number of formats as we want. If a value doesn't match against provided formats, it will return [`FormatException`](https://msdn.microsoft.com/en-us/library/system.formatexception(v=vs.110).aspx). Please refer to my whole test cases.
* For serialisation, it only considers the first format in the format list.

Here are examples of usage:

```yaml
datetime: 2016-12-31 12:34:56
```

```csharp
var builder = new DeserializerBuilder()
    .WithNamingConvention(new CamelCateNamingConvention())
    .WithTypeConverter(new DateTimeConverter(DateTimeKind.Utc, new[] { "yyyy-MM-dd HH:mm:ss" });

var deserialiser = builder.Build();

var deserislised = deserialiser.Deserialize<TestObject>(yaml);
```

```csharp
var builder = new SerializerBuilder()
    .WithNamingConvention(new CamelCateNamingConvention())
    .WithTypeConverter(new DateTimeConverter(DateTimeKind.Utc, new[] { "yyyy-MM-dd HH:mm:ss" });

var serialiser = builder.Build();

var serialised = serialiser.Serialize(deserialised);
```

**Please note**: According to [Standard Date and Time Format Strings](https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx), there's no difference between "F" and "U", in terms of string representation. Therefore, if both are provided into the converter at the same time, it will return unexpected result.

### DIFF
--- a/YamlDotNet/Serialization/Converters/DateTimeConverter.cs
+++ b/YamlDotNet/Serialization/Converters/DateTimeConverter.cs
@@ -33,19 +33,22 @@ namespace YamlDotNet.Serialization.Converters
     /// </summary>
     public class DateTimeConverter : IYamlTypeConverter
     {
-        private readonly DateTimeKind _kind;
-        private readonly string[] _formats;
+        private readonly DateTimeKind kind;
+        private readonly IFormatProvider provider;
+        private readonly string[] formats;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DateTimeConverter"/> class.
         /// </summary>
         /// <param name="kind"><see cref="DateTimeKind"/> value. Default value is <see cref="DateTimeKind.Utc"/>. <see cref="DateTimeKind.Unspecified"/> is considered as <see cref="DateTimeKind.Utc"/>.</param>
+        /// <param name="provider"><see cref="IFormatProvider"/> instance. Default value is <see cref="CultureInfo.InvariantCulture"/>.</param>
         /// <param name="formats">List of date/time formats for parsing. Default value is "<c>G</c>".</param>
         /// <remarks>On deserializing, all formats in the list are used for conversion, while on serializing, the first format in the list is used.</remarks>
-        public DateTimeConverter(DateTimeKind kind = DateTimeKind.Utc, params string[] formats)
+        public DateTimeConverter(DateTimeKind kind = DateTimeKind.Utc, IFormatProvider provider = null, params string[] formats)
         {
-            this._kind = kind == DateTimeKind.Unspecified ? DateTimeKind.Utc : kind;
-            this._formats = formats.DefaultIfEmpty("G").ToArray();
+            this.kind = kind == DateTimeKind.Unspecified ? DateTimeKind.Utc : kind;
+            this.provider = provider ?? CultureInfo.InvariantCulture;
+            this.formats = formats.DefaultIfEmpty("G").ToArray();
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


YamlDotNet/Serialization/Converters/DateTimeConverter.cs
```diff
@@ -0,0 +1,115 @@
...
+using YamlDotNet.Core.Events;
+
+namespace YamlDotNet.Serialization.Converters
+{
+    /// <summary>
+    /// This represents the YAML converter entity for <see cref="DateTime"/>.
+    /// </summary>
+    public class DateTimeConverter : IYamlTypeConverter
+    {
+        private readonly DateTimeKind _kind;
```
aaubry: We don't prefix members with underscores in this project. Please remove them to ensure consistency with the codebase.

YamlDotNet/Serialization/Converters/DateTimeConverter.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <param name="parser"><see cref="IParser"/> instance.</param>
+        /// <param name="type"><see cref="Type"/> to convert.</param>
+        /// <returns>Returns the <see cref="DateTime"/> instance converted.</returns>
+        /// <remarks>On deserializing, all formats in the list are used for conversion.</remarks>
+        public object ReadYaml(IParser parser, Type type)
+        {
+            var value = ((Scalar)parser.Current).Value;
+            var style = this._kind == DateTimeKind.Local ? DateTimeStyles.AssumeLocal : DateTimeStyles.AssumeUniversal;
+
+            var dt = DateTime.ParseExact(value, this._formats, CultureInfo.InvariantCulture, style);
```
aaubry: Should we allow to specify the culture that is to be used for reading and writing? What do you think?

YamlDotNet/Serialization/Converters/DateTimeConverter.cs
justinyoo: Oh, I overlooked that convention. Sorry about that. I'll remove those underscore prefixes.

YamlDotNet/Serialization/Converters/DateTimeConverter.cs
justinyoo: I also thought about that and concluded to keep invariant culture as I thought it should not be culture-specific.

However, as soon as I read your comment, an idea popped up on top of my head. I'll update the `DateTimeConverter` class to pass `CultureInfo` value to handle specific culture, like `DateTimeKind` value. But I'd still like to keep `CultureInfo.InvariantCulture` as its default value, if no CultureInfo is provided. How does it sound to you?

justinyoo:
@aaubry As requested, I've made some changes.

justinyoo:
@aaubry Could you run the build at AppVeyor again? It should pass all tests as I ran the test same way as what the build process did. I used `xunit.runner.console.2.2.0-beta2-build3300\tools\xunit.console.x86.exe` and all tests passed, by the way.

justinyoo:
@aaubry FYI, I've run this build against my forked branch on AppVeyor and confirm the build was successful. For some reason, the build process was interrupted at that time.

https://ci.appveyor.com/project/justinyoo/yamldotnet/build/4.0.1-pre1

Could you run the build, https://ci.appveyor.com/project/aaubry/yamldotnet/build/4.0.1-pre314, again at your end, please?

aaubry:
I was unable to reply sooner, sorry. I ran the tests locally and the test fails occasionally. But if I run the tests again, they pass. I'm not really sure what's going on, but I'll investigate further.

aaubry:
I have updated FakeItEasy to version 2.3.3 and now it seems that the tests always pass.
I'll do a release soon that will include this change. Thanks

aaubry:
Hi, @justinyoo, I noticed recently that some of the DateTime tests fail on Linux with Mono. I had to [comment out some of them](https://github.com/aaubry/YamlDotNet/pull/255/files/d23fac5b29c22437ed5877363a3c7afb9c2c88f6#diff-5d84ddaf2bd4a8969e7bd1625e6fe4e9), apparently the ones that use the Korean AM/PM designator. I thought that maybe you knew of a workaround? Thanks

justinyoo:
@aaubry ORLY? Yeah, for now you can comment that out. Seems that tests only fail where AM/PM designator exists. I'll have a look.

Issue #237, opened 2017-01-26
Strings converted to numbers


Untit1ed:
```
var swaggerDocument = new Dictionary<string, object>(){
   {"version", "2.0"}
};
var serializer = new Serializer();
var result = serializer.Serialize(swaggerDocument);
```

results in `- version: 2.0` instead of `- version "2.0"`

replaysMike:
bump for this issue.

replaysMike:
This is fixed in https://github.com/aaubry/YamlDotNet/pull/239 - but it will have to be reviewed & accepted by Antoine for accuracy - tests pass and fixes this issue.

PR #239, opened 2017-01-31
Fixes issue #237 - Adds serialization support to treat floating point numbers defined in…

replaysMike:
… strings as quoted strings.

Previously, floating point numbers were treated as normal numbers even if defined as a string in the passed in object.
Adds test case.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -300,6 +300,20 @@ namespace YamlDotNet.Test.Serialization
         }
     }
 
+    public class QuotedFloatingPoint
+    {
+        public string FloatingPointString { get; }
+        public string IntegerString { get; }
+        public double FloatingPoint { get; }
+
+        public QuotedFloatingPoint(string floatingPointString, string integerString, double floatingPoint)
+        {
+            FloatingPointString = floatingPointString;
+            IntegerString = integerString;
+            FloatingPoint = floatingPoint;
+        }
+    }
+
     public class InheritanceExample
     {
         public object SomeScalar { get; set; }
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -1152,6 +1152,15 @@ namespace YamlDotNet.Test.Serialization
             writer.ToString().Should().Contain("new_key_here: new_value");
         }
 
+        [Fact]
+        public void SerializeQuotedFloatingPointString()
+        {
+            var expectedResult = Yaml.StreamFrom("quotedFloatingPoint.yaml").ReadToEnd();
+            var testDocument = new QuotedFloatingPoint("100.25", "100", 100.75);
+            var result = Serializer.Serialize(testDocument);
+            result.Should().Be(expectedResult);
+        }
+
         [Theory, MemberData("SpecialFloats")]
         public void SpecialFloatsAreHandledCorrectly(FloatTestCase testCase)
         {
--- a/YamlDotNet.Test/YamlDotNet.Test.csproj
+++ b/YamlDotNet.Test/YamlDotNet.Test.csproj
@@ -144,6 +144,7 @@
     <EmbeddedResource Include="files\multi-doc-tag.yaml" />
     <EmbeddedResource Include="files\unicode-32bits-escape.yaml" />
     <EmbeddedResource Include="files\anchors-overwriting.yaml" />
+    <EmbeddedResource Include="files\quotedFloatingPoint.yaml" />
     <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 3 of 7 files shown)


replaysMike:
I've reviewed what I could about the Yaml spec, and in addition to some behaviors seen in other uses of Yaml. This library treats all numbers as unquoted elements even when serializing from string properties in objects passed in. I viewed this as being incorrect, and breaks swagger generated schema as their spec requires the version number to be quoted.

https://github.com/aaubry/YamlDotNet/issues/237

aaubry:
Thanks for taking the time to look into this. While I agree that the current behavior is buggy, I think that your proposed fix is too specific for floating point numbers. Also, I don't think that this should be solved directly by the parser. The specification says that the parser should produce [non-specific tags](http://www.yaml.org/spec/1.2/spec.html#tag/resolution/), that can then be resolved according to a schemas](http://www.yaml.org/spec/1.2/spec.html#Schema).

I have started working on an implementation of schemas on the [feat-schema](https://github.com/aaubry/YamlDotNet/tree/feat-schemas) branch. I am not completely satisfied with the current implementation, but you could take a look if you are interested.

Dargmuesli:
What makes you feel not completely satisfied with the current implementation? I have a similar problem with the docker compose.yml file. There too needs the version number to be enclosed in quotes. Why are those schemas not a thing yet? Can I support you in any way?

aaubry:
Regarding schemas, I feel that the whole tag resolution process needs to be reworked to comply with the specification. When parsing, the parser should assign [non-specific tags](http://www.yaml.org/spec/1.2/spec.html#tag/non-specific/) to the nodes. Then a schema could be applied to resolve such tags. I have implemented part of that, but it does not allow to use a more advanced schema, such as a .NET type to resolve the tags.

But your problem seems to have more to do with serialization, am I correct? In that case, you probably only need to add `[YamlMember(ScalarStyle = ScalarStyle.DoubleQuoted)]` to the properties that you need to be quoted. I don't think that schemas would make any difference.

Dargmuesli:
I played with serialization for a few hours now and got your suggestion working. But I don't know how this gets me any further.

I am using [powershell-yaml](https://github.com/cloudbase/powershell-yaml) which uses YamlDotNet. I pass a hashtable with all my data to the serializer. Within this hashtable all strings that should be strings are strings and all numbers that should be numbers are numbers. What YamlDotNet outputs is: numbers are numbers, strings - that contain numbers - are numbers.

The strings I give YamlDotNet are strings on purpose! Even if they contain only numbers. I cannot work with the current output, because I need strings to be strings, no matter what their content is!

How can I achieve that?

aaubry:
The thing is that your strings are being serialized as strings, but in YAML, strings are usually not quoted, unless they contain special characters. But if the parser interprets them as a number, that will indeed be a problem. This has convinced me that schemas are also necessary when emitting documents. I will have to reflect more on this.

As a workaround, if you are using a sufficiently recent version of YamlDotNet, you can try to substitute your strings with instances of YamlScalarNode. This will allow you to specify the scalar style. I am away from the computer at the moment, but I can try to give you an example later.

Dargmuesli:
If you could give me an example, that would be really great!

aaubry:
I checked powershell-yaml, and it is using a [really old version of YamlDotNet](https://github.com/aaubry/YamlDotNet/releases/tag/v2.0.0). This means that my suggestion won't work, and also that even if the schemas are implemented, you probably won't be able to use them from that wrapper.

I also tried creating a class in Powershell that implemented `IYamlSerializable`, but that fails because powershell-yaml seems to be doing some kind of conversion of the objects that it receives, which causes an exception.

You will probably be better off if you use YamlDotNet directly. Just install the nuget package or extract  YamlDotNet.dll somewhere and load it:
```powershell
Add-Type -Path './lib/YamlDotNet.dll'

function New-DoubleQuotedString($value)
{
    $scalar = [YamlDotNet.RepresentationModel.YamlScalarNode]::new($value)
    $scalar.Style = [YamlDotNet.Core.ScalarStyle]::DoubleQuoted
    return $scalar
}


$serializer = [YamlDotNet.Serialization.SerializerBuilder]::new().Build()

$serializer.Serialize(@{"hello"="world"; "quoted"=New-DoubleQuotedString("bar")"anArray"=@(1,2,3); "nested"=@{"array"=@("this", "is", "an", "array")}})
```
This gives the following output (notice the double quotes):
```yaml
hello: world
quoted: "bar"
nested:
  array:
  - this
  - is
  - an
  - array
```

Dargmuesli:
That's a great example, thank you so much! I'll ditch `powershell-yaml` and use your implementation. Thank you once more, it would have taken me very long to figure this out.

julealgon:
@aaubry , what's the state of this change? Is is still a problem in the latest versions, or has it been fixed some other way (and if so, in which version)? Is the change from @replaysMike still needed?

EdwardCooke:
I believe this has been fixed in version 12.x. I would give the latest version a try. There were a lot of special strings that needed to be quoted for it to work.
This PR also appears to be abandoned, it's over a year old. I'm going to close it.

PR #244, opened 2017-03-29
Port to .NET Standard

nbarbettini:
Hey @aaubry, I took another stab at porting this library to .NET Standard. The tooling in VS2017 is updated now, and it no longer requires adding an xproj/project.json file. This PR should supercede #210 and #240.

I tried to keep the changes as minimal as possible so everything else will still work (fingers crossed).

Like before, the library targets .NET Framework 3.5. However, it also now targets .NET Standard 1.3 (.NET Standard is basically an evolution of the idea of PCLs, see [here](https://docs.microsoft.com/en-us/dotnet/articles/standard/library)). All the tests pass, and I think all the other build profiles shouldn't be affected.

The only todo here is fixing the assembly versioning. I think it's being added by the Cake build script? I didn't quite follow it.

PR #255, opened 2017-05-08
Port to .NET Standard

aaubry:
This contains all the changes proposed and discussed in #244.

What's included:
* Added two targets, `Release-DotNetStandard-Unsigned` and `Release-DotNetStandard-Signed` that produce assemblies targeting [`netstandard1.3`](https://docs.microsoft.com/en-us/dotnet/articles/standard/frameworks).
* Added the `netstandard1.3` assemblies to the Nuget packages. This means that using YamlDotNet on .NET core should no longer be a problem.
* Fixed some issues that surfaced when updating the AppVeyor image to VS2017 because it already includes the Unity prerequisites.
* Added a Dockerfile that builds an image (`aaubry/yamldotnet`) that contains all the dependencies needed to build YamlDotNet on Linux. This was necessary to overcome some difficulties with mono, but it is also useful for anyone wanting to build the library without having to install anything (besides Docker).
* Changed the Travis build so that it uses our docker image and performs a full build, including unit tests. This will help detect problems in case a test passes on Windows only.

What's still missing:
* The unit tests project should also be converted so that tests are also run against the .NET Standard assemblies.


YamlDotNet/Properties/AssemblyInfo.cs
```diff
@@ -1,66 +1,66 @@
...
+// Version information for an assembly consists of the following four values:
+//
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Revision and Build Numbers
+// by using the '*' as shown below:
+[assembly: AssemblyVersion("4.1.1.0")]
```
conniey: I am not sure if you meant to check this in since it's always updated during testing. :)

YamlDotNet.Test/Serialization/DateTimeConverterTests.cs
```diff
@@ -348,18 +348,18 @@ public void Given_Yaml_WithTimeFormat_ReadYaml_ShouldReturn_Result(string format
         [InlineData("Y", "fr-FR", "janvier 2017")]
         [InlineData("d", "ko-KR", "2017-01-13")]
         [InlineData("D", "ko-KR", "2017년 1월 13일 금요일")]
-        [InlineData("f", "ko-KR", "2017년 1월 13일 금요일 오전 5:32")]
-        [InlineData("F", "ko-KR", "2017년 1월 13일 금요일 오전 5:32:06")]
-        [InlineData("g", "ko-KR", "2017-01-13 오전 5:32")]
-        [InlineData("G", "ko-KR", "2017-01-13 오전 5:32:06")]
+        // [InlineData("f", "ko-KR", "2017년 1월 13일 금요일 오전 5:32")]
```
conniey: I tried to look into these failures because the tests work on Windows but not Linux.  From what I could tell, Mono is having troubles trying to fetch the AM/PM designator for ko-KR (because it is always null).

conniey:
LGTM. 👍 Thanks a lot for putting this together!

YamlDotNet.Test/Serialization/DateTimeConverterTests.cs
aaubry: I left a comment to @justinyoo , who contributed this code. Maybe he knows more about this problem.

justinyoo: @conniey @aaubry Yep, I'll have a look. In the meantime, you can comment out those lines.

nbarbettini:
Confirmed that 4.2.0 works fine in my .NET Standard 1.3 project. I don't have to use `PackageTargetFallback` anymore! 🎉 Thanks @aaubry and @conniey.

PR #257, opened 2017-05-11
Migrate tests to .NET Core

conniey:
* This is a PR on top of your current PR: #255 
* Migrates existing tests to netcoreapp1.0
* Updates build steps in build.cake

Sample builds: 
* AppVeyor: https://ci.appveyor.com/project/conniey/yamldotnet/build/0.0.42
* TravisCI: https://travis-ci.org/conniey/YamlDotNet/builds/231319566

Currently 1 of the tests is failing on travis-ci on .NETCoreApp1.0. Looking into it.
```
Starting test execution, please wait...
[xUnit.net 00:00:00.8345615]   Discovering: YamlDotNet.Test
[xUnit.net 00:00:01.4937744]   Discovered:  YamlDotNet.Test
[xUnit.net 00:00:01.6688416]   Starting:    YamlDotNet.Test
[xUnit.net 00:00:03.1973528]     YamlDotNet.Test.Serialization.DateTimeConverterTests.Given_Yaml_WithLocaleAndTimeFormat_ReadYaml_ShouldReturn_Result(format: "d", locale: "ko-KR", value: "2017-01-13") [FAIL]
[xUnit.net 00:00:03.1993405]       System.FormatException : String was not recognized as a valid DateTime.
[xUnit.net 00:00:03.2043744]       Stack Trace:
[xUnit.net 00:00:03.2075158]            at System.DateTimeParse.ParseExact(String s, String format, DateTimeFormatInfo dtfi, DateTimeStyles style)
[xUnit.net 00:00:03.2087634]         /build/YamlDotNet.Test/Serialization/DateTimeConverterTests.cs(366,0): at YamlDotNet.Test.Serialization.DateTimeConverterTests.Given_Yaml_WithLocaleAndTimeFormat_ReadYaml_ShouldReturn_Result(String format, String locale, String value)
Failed   YamlDotNet.Test.Serialization.DateTimeConverterTests.Given_Yaml_WithLocaleAndTimeFormat_ReadYaml_ShouldReturn_Result(format: "d", locale: "ko-KR", value: "2017-01-13")
Error Message:
 System.FormatException : String was not recognized as a valid DateTime.
Stack Trace:
   at System.DateTimeParse.ParseExact(String s, String format, DateTimeFormatInfo dtfi, DateTimeStyles style)
   at YamlDotNet.Test.Serialization.DateTimeConverterTests.Given_Yaml_WithLocaleAndTimeFormat_ReadYaml_ShouldReturn_Result(String format, String locale, String value) in /build/YamlDotNet.Test/Serialization/DateTimeConverterTests.cs:line 366
[xUnit.net 00:00:04.4661091]   Finished:    YamlDotNet.Test
```

### DIFF
--- a/YamlDotNet.Test/Serialization/DateTimeConverterTests.cs
+++ b/YamlDotNet.Test/Serialization/DateTimeConverterTests.cs
@@ -346,7 +346,7 @@ namespace YamlDotNet.Test.Serialization
         [InlineData("u", "fr-FR", "2017-01-13 05:25:08Z")]
         [InlineData("U", "fr-FR", "vendredi 13 janvier 2017 05:25:08")]
         [InlineData("Y", "fr-FR", "janvier 2017")]
-        [InlineData("d", "ko-KR", "2017-01-13")]
+        // [InlineData("d", "ko-KR", "2017-01-13")]
         [InlineData("D", "ko-KR", "2017년 1월 13일 금요일")]
         // [InlineData("f", "ko-KR", "2017년 1월 13일 금요일 오전 5:32")]
         // [InlineData("F", "ko-KR", "2017년 1월 13일 금요일 오전 5:32:06")]


aaubry:
Regarding the failing tests, it seems that there is a bug somewhere in the framework. Consider the following program:

```c#
using System;
using System.Globalization;
					
public class Program
{
	public static void Main()
	{
		var culture = new CultureInfo("ko-KR");
		Console.WriteLine(DateTime.Now.ToString("d", culture));
	}
}
```
If I run it on .NET 4.5, or on Mono, the output is
```
2017-05-12
```

But if I run it on .NET Core, the output is
```
17. 5. 12.
```

I don't know which is correct, but they cannot be both right.

conniey:
I've created an issue in https://github.com/dotnet/corefx/issues/19711. Hopefully tarek respond to the issue soon.

conniey:
@aaubry  I did some digging and this behaviour is by-design.  .NET Core gets its culture data from the underlying framework.  On Windows, it is [National Language Support (NLS)](https://msdn.microsoft.com/en-us/library/windows/desktop/dd319078(v=vs.85).aspx).  Most Linux distributions have ICU installed, which obtains its globalization data from CLDR.  Mono also uses NLS, which is why you see the same behaviour. 

ICU Korean Culture info: http://www.localeplanet.com/icu/ko-KR/

See https://github.com/dotnet/coreclr/issues/1734#issuecomment-146882541 for more insight.

One option to fix this test is to use `RuntimeInformation.IsOSPlatform` on .NET Core for the tests.

>In short, until Windows and CLDR are in 100% agreement, there will always be differences between the culture data on Windows and Linux. Which means if you are dependent on this data (like your tests are), then you will have to switch depending on what platform you are running on. In our tests, we use RuntimeInformation.IsOSPlatform(OSPlatform.Windows) or Linux or OSX, etc.

build.ps1
```diff
@@ -185,5 +185,5 @@ if (!(Test-Path $CAKE_EXE)) {
 
 # Start Cake
 Write-Host "Running build script..."
-Invoke-Expression "& `"$CAKE_EXE`" `"$Script`" -target=`"$Target`" -configuration=`"$Configuration`" -verbosity=`"$Verbosity`" $UseMono $UseDryRun $UseExperimental $ScriptArgs"
+Invoke-Expression "& `"$CAKE_EXE`" `"$Script`" -target=`"$Target`" -configuration=`"$Configuration`" -buildVerbosity=`"$Verbosity`" $UseMono $UseDryRun $UseExperimental $ScriptArgs"
```
aaubry: This argument should be `verbosity` and not `buildVerbosity`. `verbosity` controls the verbosity of Cake, and `buildVerbosity` is an argument that I added myself to control the verbosity of msbuild, because I found it useful to be able to control them separately.

aaubry:
Ok, thanks for porting the tests, and for looking into this issue. I don't think that these tests are that important anyway. Let's just comment out the ones that fail, and I'll see later if they can be reworked.

conniey:
Alright. I commented out that one test and the travis-ci build works now. https://travis-ci.org/conniey/YamlDotNet/builds/231735514

aaubry:
I made [some additional changes](https://github.com/aaubry/YamlDotNet/commit/3a762632b45e8c6283297e65e82b2160a512f334) so that the unit test-related tasks always use the same method and, with the test disabled, [everything builds](https://ci.appveyor.com/project/aaubry/yamldotnet/build/4.1.1-PullRequest0406/tests) [everywhere](https://travis-ci.org/aaubry/YamlDotNet/builds/231741240). I will merge everything into master and release a new version.
Again, **thanks for your help!**

PR #278, opened 2017-09-06
[WIP] handling exclamation mark characters when parsing YAML

shiftkey:
I wanted to spike up a proof-of-concept which required consuming YAML files generated by the [vcr](https://rubygems.org/gems/vcr) gem, but encountered a parsing problem with this style of node:

```yaml
      Etag:
      - ! '"24ec5-1b6-4059a80bfd280"'
```

It looks like this is a known part of the [1.2 spec](http://www.yaml.org/spec/1.2/spec.html#id2784064):

> By explicitly specifying a “!” non-specific tag property, the node would then be resolved to a “vanilla” sequence, mapping, or string, according to its kind.

And this SO thread mentions that it's a known issue with YamlDotNet: https://stackoverflow.com/questions/9664113/what-does-a-single-exclamation-mark-do-in-yaml

I'm honestly not sure if this is intended to be supported, or even how to make it work properly, but I figure I'd start this off by adding a failing test and see where to go from there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ bin
 obj
 tools
 
+# Visual Studio local cache
+.vs/
+
+
 # mstest test results
 TestResults
 *.suo
--- a/YamlDotNet.Test/RepresentationModel/YamlStreamTests.cs
+++ b/YamlDotNet.Test/RepresentationModel/YamlStreamTests.cs
@@ -104,6 +104,12 @@ namespace YamlDotNet.Test.RepresentationModel
         }
 
         [Fact]
+        public void RoundTripVCRExample()
+        {
+            RoundtripTest("vcr-example-file.yaml");
+        }
+
+        [Fact]
         public void RoundtripExample1()
         {
             RoundtripTest("01-directives.yaml");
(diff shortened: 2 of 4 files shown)


shiftkey:
Not sure what's up with CI, but here's the details of the test error I'm seeing locally:

```
  Starting:    YamlDotNet.Test
    YamlDotNet.Test.RepresentationModel.YamlStreamTests.RoundTripVCRExample [FAIL]
      YamlDotNet.Core.SyntaxErrorException : (Line: 18, Col: 9, Idx: 315) - (Line: 18, Col: 10, Idx: 316): While parsing a tag, did not find expected tag URI.
      Stack Trace:
        YamlDotNet\Core\Scanner.cs(2163,0): at YamlDotNet.Core.Scanner.ScanTagUri(String head, Mark start)
        YamlDotNet\Core\Scanner.cs(1264,0): at YamlDotNet.Core.Scanner.ScanTag()
        YamlDotNet\Core\Scanner.cs(1202,0): at YamlDotNet.Core.Scanner.FetchTag()
        YamlDotNet\Core\Scanner.cs(415,0): at YamlDotNet.Core.Scanner.FetchNextToken()
        YamlDotNet\Core\Scanner.cs(212,0): at YamlDotNet.Core.Scanner.FetchMoreTokens()
        YamlDotNet\Core\Scanner.cs(128,0): at YamlDotNet.Core.Scanner.MoveNextWithoutConsuming()
        YamlDotNet\Core\Parser.cs(51,0): at YamlDotNet.Core.Parser.GetCurrentToken()
```

shiftkey:
Going to try a different angle to this proof of concept before I come back to this.

aaubry:
The parsing of tags definitely needs some work. I'm not sure why this fails in this case, but this should be fixed.
Regarding CI, it seems that it is broken. I'll look into it.

PR #279, opened 2017-09-20
Add travis/appveyor/nuget badges

damageboy:


aaubry:
Thanks. I need to check the build and see why it is failing.

Issue #259, opened 2017-05-16
Two level <<


tauvinen:
Simplified example:

```
base: &level1
  X: "X"

base: &level2
  <<: *level1
  Y: "Y"

ok:
  <<: *level1

fail:
  <<: *level2
```

And basic usage like:
```csharp
var deserializer = new DeserializerBuilder().
                  .Build();
var parser = new MergingParser(new Parser(fileReader));
var confFile = deserializer.Deserialize<xxxx>(parser);

```

Deserialize results in exception: "Property '<<' not found on type 'XXX.base'". If "fail" reference is changed to "*level1" it works.

I am not certain if this should work, but would be logical?

aaubry:
I would expect this to work, but the `MergingParser` is a bit of a hack, and maybe it does not handle recursivity. I will look into this when I can.

PR #283, opened 2017-10-22
Fix issue #259 - Improve nested reference merging

vCipher:
Hi!

This fixes issue #259. My solution is to place all of the parsing events to LinkedList and insert referenced content after AnchorAlias. It allows to resolve all references (also nested) in one pass. Also i've used reference map to avoid anchor lookup every time.


aaubry:
This looks promising! I'll need a few days to properly review this.
Thanks

aaubry:
This is much cleaner than the previous implementation, and also supports nesting, which is excellent. Thanks!

Issue #293, opened 2017-12-13
Can not deserialize DateTimeOffset in .netcoreapp2.0


vCipher:
Simple example:
```
class Program
{
    static void Main(string[] args)
    {
        var deserializer = new Deserializer();
        var yaml = "before: 2015-10-23T12:23:15";
        var config = deserializer.Deserialize<Config>(yaml);
    }
}

class Config
{
    [YamlMember(Alias = "before")]
    public DateTimeOffset Before { get; set; }
}
```

Rises exception:
```
System.InvalidCastException: Invalid cast from 'System.String' to 'System.DateTimeOffset'.
   at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
   at System.String.System.IConvertible.ToType(Type type, IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at YamlDotNet.Serialization.Utilities.TypeConverter.ChangeType(Object value, Type destinationType, CultureInfo culture)
   at YamlDotNet.Serialization.Utilities.TypeConverter.ChangeType(Object value, Type destinationType)
   at YamlDotNet.Serialization.NodeDeserializers.ScalarNodeDeserializer.YamlDotNet.Serialization.INodeDeserializer.Deserialize(IParser parser, Type expectedType, Func`3 nestedObjectDeserializer, Object& value)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
```

It think that this exception occurs because of getting ```public static T Parse(string, IFormatProvider)``` fails in https://github.com/aaubry/YamlDotNet/blob/a724846c55dbc601acbca694d5f6ab3f2b2f3459/YamlDotNet/Serialization/Utilities/TypeConverter.cs#L226

This getting of Parse method fails because of ```Suspicious comparison: there is no type in the solution which is inherited from both 'System.Reflection.ParameterInfo' and 'System.Type'``` https://github.com/aaubry/YamlDotNet/blob/b758a2961bdb0b042f7f94e1fe6cddbb34756402/YamlDotNet/Helpers/Portability.cs#L247

aaubry:
Humm, this looks like a bug. The code should probably be this instead:
```c#
[...]

PR #294, opened 2017-12-17
Fixes issue #293 - Can not deserialize DateTimeOffset in .netcoreapp2.0

vCipher:
- Change comparation scheme in ```ReflectionExtentions::GetPublicStaticMethod```
- Adds test case


aaubry:
Thanks for this fix. I changed the test slightly so that it runs on every target.
I am doing a new release, 4.2.3 that will include this fix.

PR #302, opened 2018-02-12
Avoid definition of SerializationException on NET Standard

jhickson:
# Problem

When a project targeting NET Standard 2.0 pulls in YamlDotNet 4.3.0 then
build errors related to the multiple definition of `System.Runtime.Serialization.SerializationException` occur. This is because the NET Standard (1.3) version of YamlDotNet provides its own definition of this type, but NET Standard reintroduced it in version 2.0.

# Solution

The problem is easily fixed by removing the local definition and instead referencing `System.Runtime.Serialization.Primitives` which defines the type for pre-NET Standard 2.0.

### DIFF
--- a/YamlDotNet/Core/Emitter.cs
+++ b/YamlDotNet/Core/Emitter.cs
@@ -534,9 +534,8 @@ namespace YamlDotNet.Core
         private bool IsUnicode(Encoding encoding)
         {
             return encoding is UTF8Encoding ||
-                encoding is UnicodeEncoding ||
-                encoding is UTF7Encoding ||
-                encoding is UTF8Encoding;
+                   encoding is UnicodeEncoding ||
+                   encoding is UTF7Encoding;
         }
 
         private void AnalyzeTag(string tag)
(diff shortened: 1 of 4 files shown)


jhickson:
@aaubry I'm not sure why the Travis build is failing. Judging by the error message, I don't think it's due to my change.

jhickson:
@aaubry I realised there were potentially other problems on NET Standard with using fake definitions of `SerializableAttribute`, `System.DBNull`, `System.TypeCode`, and `System.Text.UTF7Encoding`. All these are available as platform extensions for NET Standard 1.3 and are part of NET Standard 2.0.

aaubry:
Thanks for this. I'll try to review your PR asap.
It makes sense that these types are available in netstandard. This code was intended for the old portable class library target, and was not updated.
Do not worry about the Travis build, the build fails on pull requests. I think that is a problem with the Git Version configuration, but I did not have time to look into it.

aaubry:
I've removed the reference to System.Data.Common and instead used the type name to detect DBNull, to avoid that dependency from being pushed to everybody, and also because we do not really use the type anyway.

jhickson:
@aaubry Thanks for reviewing and merging. I see you've released it already too.

aaubry:
👍

Issue #117, opened 2015-01-30
Serializing also fields and not just properties will make life easier


oscarsan:
I have an small problem, I'm using structures for 2 purposes, deserialize and serialize from Yaml files but also to download it to a realtime system which needs Marshaled variables. So for example

```
public struct Configuration
    {
        public NameValue Name;
        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 2)]
        public BoolValue[] AllowedInDirection; 
   }
```

In this case the BoolValue is something that I need to Marshal in order to been able to download it to the final component. Converting this in a property in order to been able to use the Deserializer and Serializer is quite big work, I will need to create different structures for the realtime component and for the YAML file and that is not go.

So, why not to make this marvelous component so that it will be also possible to serialize a deserialize fields and not just properties inside a class or structure.

svermeulen:
This would also play a lot more nicely with Unity since Unity serialization does the opposite (requires that serializable data be in fields and not in properties)

zatherz:
Bump

FullStackForger:
+1 from me here.  YamlDotNet is awesome but making it work with unity generates lots of unnecessary work.

PathogenDavid:
It is worth pointing out that this is preventing serialization of ValueTuple since they use public fields.

aaubry:
This has been added to [release 5.4.0](https://github.com/aaubry/YamlDotNet/releases/tag/v5.4.0).

PR #306, opened 2018-03-16
Add support for deserializing fields in addition to properties

thisisthedave:
I made changes so that I can deserialize yaml to fields, which should address #117 "Serializing also fields and not just properties will make life easier". This is convenient for Unity objects with [SerializeField] members because Unity disallows using properties for these fields. 

I'm not setup to test on platforms other than net46, so there may be issues on other platforms. I added a simple unit test. I'll attempt to execute the docker-based build/test soon when I have a chance to review how it works. I haven't done a pull request on a public repo before, so let me know if there is protocol I'm overlooking.

Thank you for publishing this library! It has been very useful to us.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -347,6 +347,11 @@ namespace YamlDotNet.Test.Serialization
         public string aaa { get; set; }
     }
 
+    public class SimpleField
+    {
+        public string theField;
+    }
+
     public class SimpleScratch
     {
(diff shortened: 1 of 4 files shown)


kasper3:
@aaubry, could this be considered for the next version of YamlDotNet?

aaubry:
Apparently this PR got lost somehow and in the meantime serialization of fields was independently added  to [release 5.4.0](https://github.com/aaubry/YamlDotNet/releases/tag/v5.4.0). Can this be closed, or is there a feature here that is not provided by the current implementation ?

PR #309, opened 2018-03-24
small change to comment parsing logic

### LINKED ISSUE #308: Comment on line start parsed as `inline`
After literal or folded block standalone comment parsed as inline. 
For example:
```yaml
|-
  text
#comment
```
will translate to `literalScalar("text")` and `inlineComment("comment")`.

With "-" chomping it leads to a problem - if emit parsed result as is, resulting file will be:
```yaml
|-
  text #comment
```

lexarchik:
trying to fix #308


aaubry:
This looks good. Thanks

PR #313, opened 2018-04-05
Update Scanner.cs to correct spelling errors

thoughtcroft:
Correct multiple instances of "indentation" incorrectly spelled as "intendation"


aaubry:
Thanks for taking the time to correct this!

PR #345, opened 2018-08-29
Avoid ArgumentException (Key already exists) in parallelized usage.

sbhenderson:
Hi there.

This PR is to solve an issue I have found in my own usage where, during parallelized serializations using the same Serializer, the CachedTypeInspector would occasionally throw on the Add due to the key already existing (because a different thread already cached it). By using the indexer, we avoid this exception as the item will be added or modified as necessary.

It is clear that this is more of a band-aid than a full solution, so I am willing to hear alternatives. Some examples:
- We could use a lock here instead. 
- Use ConcurrentDictionary where available. I have already prepared code to use it on applicable platforms, and this could be combined with the lock on non-applicable platforms.
- Check for existence of the key once more before adding, but this simply minimizes the window of opportunity for parallelization to mess it up.

I did read the guidelines, and I recognize I am not also contributing a test. I wanted to ask whether or not this change deserved it. Unfortunately, anything with parallelization/threading is difficult to get reliable tests, but if you feel that it's worth it, let me know.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,13 @@ Please read [CONTRIBUTING.md](CONTRIBUTING.md) for guidelines.
 
 # Changelog
 
+## Version 5.4.0
+
+New features:
+* **Enable serialization of public fields.**  
+  YamlDotNet will now also serialize public fields. This feature is enabled by default,
+  but it can be disabled by calling `IgnoreFields()` no the `SerializerBuilder` or `DeserializerBuilder`.
+
 ## Version 5.3.1
 
 New features:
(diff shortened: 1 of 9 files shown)


aaubry:
I've only been able to look at this now. This will need a better fix, as the current implementation, even with your changes, is not thread-safe.

sbhenderson:
@aaubry So would you like to use the ConcurrentDictionary on platforms where available and appropriate sync locks otherwise?

aaubry:
Yes, I think that would be the best option. We also need to check if there are other code paths that are not thread-safe. I don't think there should be many because the Serializer and Deserializer are designed to be immutable, but we should double check.

PR #347, opened 2018-09-09
Fix roundtripping of tags on sequences

szilvaa:
Tags get dropped at the beginning of a sequence when serializing. This PR fixes this. Yaml tags are used extensively by AWS CloudFormation see for the [Join function](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-join.html) for example.

I used string.IsNullOrEmtpy to determine the value of isImplicit because this is what's used during parsing [here](https://github.com/aaubry/YamlDotNet/blob/5ec2dd75704d6dced41c0d49ff5528f3797d1068/YamlDotNet/Core/Parser.cs#L485).


szilvaa:
@aaubry Can you take a look please?

aaubry:
Sorry, I did not have much time to look at this. I'll publish release 5.2.1 with this fix.

szilvaa:
Thank you!!!

Issue #350, opened 2018-09-20
No interfaces for serializers?


grofit:
I have a scenario where I want to convert some json to yaml and this looks like the easiest path, but one worrying thing is that there is no interface for `Serializer` or `Deserializer` so it is difficult to mock, is there any reason why there is no `IYamlSerializer` or `IYamlDeseralizer` or anything which those classes implement so we can then just use that as the contract and mock the `Serializer` dependencies for unit testing etc?

aaubry:
There's no particular reason for that. If you want to extract the interfaces and submit a pull request, I will publish an updated package immediately.

grofit:
Closing this as the merge has added interfaces.

PR #351, opened 2018-09-21
Updated to add interfaces for serializers

grofit:
This was requested as per #350 

I have run tests and all passed apart from one which seemed to be in some pragma and I was told a dll could not be found, but may be down to me being on a different .net core version, so I just temporarily ran it as .net core 2 rather than 1 (dont worry I didnt commit that change up).


aaubry:
Thanks. I've merged the changes and released version 5.1.0, which should be available on NuGet in a few minutes.

PR #369, opened 2018-11-29
Add conditional serialization functionality

fekadeabdejene:
The main IDEA, like other JSON serialization frameworks, is to be able to serialize on some condition based in a class vs just the serialization decorators. Controlling serialization via a function in the host class which can return true or false give one or more conditions.

### DIFF
--- a/YamlDotNet.Test/Serialization/ConditionalSerializeTests.cs
+++ b/YamlDotNet.Test/Serialization/ConditionalSerializeTests.cs
@@ -1,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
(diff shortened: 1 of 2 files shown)


aaubry:
I'm not sure about this. The use case for this seems very specific, and as the PR clearly shows, it is easy to add without changing the existing code. Maybe this would be better as a separate package ?

fekadeabdejene:
Well. It up to you.  Allowing a function to determine when to serialize a property will allow flexibility. I.E. people can clean up their serialized yaml files from default or null values or any conditions they want with ease. The only choice at the moment is binary including or excluding a property for serialization unless they implement their own solution. 

As you said, in can be added with ease. So maybe it is best as a helper function in the builder if people want this functionality present.

It could also just be an example usage.

aaubry:
I will close this PR as there has I'm still not convinced that this should be built-in. Thanks

PR #386, opened 2019-02-24
CachedTypeInspector ConcurrentDictionary or locks for parallelized usage

sbhenderson:
Hi again. This is related to the PR #345.

I have gone back and decided to follow up based on what you suggested:

1. On platforms where ConcurrentDictionary is available, just use it directly without any further worry. Obviously, it's possible for multiple threads to concurrently evaluate the GetProperties for the same type (somewhat wasteful), but this impact would be significant only for larger types. I'd expect that if people had such large types, they may have considered parallelization before and will be helped by this anyway.
2. On platforms where ConcurrentDictionary isn't available, I do a simple lock on the type with an extra check on existence before evaluation. Because reads aren't thread-safe in the standard Dictionary, putting an extra check behind a lock should guarantee that it doesn't exist before attempting to evaluate. 

There is a small performance penalty to using ConcurrentDictionary or a lock in pure singlethreaded scenarios, so I wonder if we should add this as an option in the Builder classes. Thoughts?

Finally, I did a quick check of all uses of Dictionaries in the project, and the only one that seemed suspicious to me was SerializerState. However, because it's only being used for a single type right now, I don't see it as being problematic. Obviously, dictionaries are used in more places, but my quick read suggested that they were related to a single document serialization/deserialization and thus not a threat.

### DIFF
--- a/YamlDotNet/Serialization/TypeInspectors/CachedTypeInspector.cs
+++ b/YamlDotNet/Serialization/TypeInspectors/CachedTypeInspector.cs
@@ -18,10 +18,11 @@
 //  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 //  SOFTWARE.
-
+#if !(NET20 || NET35)
+#define USE_CONCURRENT_DICTIONARY
+#endif
 using System;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace YamlDotNet.Serialization.TypeInspectors
 {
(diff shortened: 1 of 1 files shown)


aaubry:
I would prefer that the platform-specific code stayed in Portability.cs as much as possible. I suggest that you create a class named `ConcurrentDictionary` in the `System.Collections.Concurrent` namespace on platforms that do not already have it. The implementation can be simple and rely on a lock, and only the methods that are actually used need to be provided.

Regarding the performance impact, we would need to measure the cost of the concurrent dictionary before adding an option to revert to the unsafe version. Let's first make the above change and make it build on all platforms.

aaubry:
This change has finally been merged into master. It was implemented in a different commit because I needed to move the code to the Portability.cs file. I will close the pull request as it no longer applies. Thanks!

rooonikz:
Hi, when can we get this fix ready in nuget? Thanks

aaubry:
I thought it was already released, but apparently I was mistaken. I'll do a release today.

rooonikz:
> I thought it was already released, but apparently I was mistaken. I'll do a release today.

Thank you for the quick response! Waiting for it

aaubry:
YamlDotNet 8.1.0 has just been uploaded to NuGet.org. It should be available in a few minutes.

rooonikz:
@aaubry perfect, thanks a lot. We are already on it!

PR #389, opened 2019-02-28
Typos and nameof

sungam3r:
Corrects a number of typos. Use nameof operator instead of string literals. In two places, the literals did not match the names of the input parameters.

### DIFF
--- a/YamlDotNet/RepresentationModel/YamlNode.cs
+++ b/YamlDotNet/RepresentationModel/YamlNode.cs
@@ -155,21 +155,7 @@ namespace YamlDotNet.RepresentationModel
         /// <summary>
         /// Gets a value indicating whether two objects are equal.
         /// </summary>
-        protected static bool SafeEquals(object first, object second)
-        {
-            if (first != null)
-            {
-                return first.Equals(second);
-            }
-            else if (second != null)
-            {
-                return second.Equals(first);
-            }
-            else
-            {
-                return true;
-            }
-        }
+        protected static bool SafeEquals(object first, object second) => object.Equals(first, second);
 
         /// <summary>
         /// Serves as a hash function for a particular type.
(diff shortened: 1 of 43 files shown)


YamlDotNet/RepresentationModel/YamlNode.cs
```diff
@@ -102,7 +102,7 @@ static internal YamlNode ParseNode(IParser parser, DocumentLoadingState state)
                 return state.GetNode(alias.Value, false, alias.Start, alias.End) ?? new YamlAliasNode(alias.Value);
             }
 
-            throw new ArgumentException("The current event is of an unsupported type.", "events");
+            throw new ArgumentException("The current event is of an unsupported type.", nameof(parser));
```
sungam3r: There is no parameter named "events". It seemed to me correct to pass parser, because it was his state condition that led to the generation of the exception.

YamlDotNet/RepresentationModel/YamlNode.cs
```diff
@@ -155,21 +155,7 @@ protected bool Equals(YamlNode other)
...
-            else if (second != null)
-            {
-                return second.Equals(first);
-            }
-            else
-            {
-                return true;
-            }
-        }
+        protected static bool SafeEquals(object first, object second) => object.Equals(first, second);
```
sungam3r: SafeEquals method does exactly the same thing as [this method](https://referencesource.microsoft.com/#mscorlib/system/object.cs,68). I did not change all the uses of the method so that there are fewer changes.

YamlDotNet/Serialization/Deserializer.cs
```diff
@@ -131,12 +131,12 @@ public object Deserialize(IParser parser, Type type)
         {
             if (parser == null)
             {
-                throw new ArgumentNullException("reader");
+                throw new ArgumentNullException(nameof(parser));
```
sungam3r: With nameof operator, there will be no more such misunderstandings.

aaubry: agreed

aaubry:
Thanks for cleaning this up! This must have taken a lot of work.

PR #396, opened 2019-04-15
Fix missing string quotes around json serialized enums (#146)

### LINKED ISSUE #146: Serializing enum values to json
When I serialize a class containing an enum value to json, it ends up with the enum name with no quotes, which results in an invalid json file. I'd like it to serialize as a string, how would I go about doing that? I've used [YamlMember(typeof(string))] on all my enum values for now, but would ideally like to be able to serialize all enum values to string, since adding this property is easily forgotten when adding new class members.

``` csharp

public class Program
{
    public enum Primitive {Sphere, Cube}
    public Primitive type {get; set;}

    public static void Main()
    {
        var serializer = new Serializer(SerializationOptions.EmitDefaults | SerializationOptions.JsonCompatible);
        serializer.Serialize(System.Console.Out, new Program());
    }
}
```

The code above serializes to this:

``` json
{
  "type": Sphere
}
```

I would like it to serialize to this:

``` json
{
  "type": "Sphere"
}
```

KapuKapu:
Fixes #146 

Json allows only [String or Number](https://www.json.org/) as Enum serialization. I suggest serialization as String to enable cleaner handling in JS code and better readability.


YamlDotNet/Serialization/EventEmitters/JsonEventEmitter.cs
```diff
@@ -53,9 +53,21 @@ public override void Emit(ScalarEventInfo eventInfo, IEmitter emitter)
...
+                        eventInfo.RenderedValue = YamlFormatter.FormatNumber(eventInfo.Source.Value);
+                    }
+                    break;
+
                 case TypeCode.Byte:
                 case TypeCode.Int16:
-                case TypeCode.Int32:
                 case TypeCode.Int64:
                 case TypeCode.SByte:
                 case TypeCode.UInt16:
```
am11: Enums can be derived from any integer types. Perhaps, it could be something like https://github.com/am11/YamlDotNet/commit/d47770c3080518edccf50e48f8b0c0cc8fa97439?

YamlDotNet.Test/Serialization/SerializationTests.cs
```diff
@@ -814,6 +814,17 @@ public void DeserializationOfDefaultsWorkInJson()
             result.MyString.Should().BeNull();
```
am11: Are the file mode changes (`100644->100755`) unintentional? To revert:

```sh
git update-index --chmod=-x YamlDotNet/Serialization/EventEmitters/JsonEventEmitter.cs YamlDotNet.Test/Serialization/SerializationTests.cs
```

YamlDotNet/Serialization/EventEmitters/JsonEventEmitter.cs
KapuKapu: Good catch, I will fix it.

YamlDotNet.Test/Serialization/SerializationTests.cs
KapuKapu: Yes, it was unintentional.

YamlDotNet.Test/Serialization/SerializationTestHelper.cs
```diff
@@ -141,6 +141,15 @@ public enum EnumExample
         Two
     }
 
+    public enum SByteEnum : sbyte { Default, Sbyte }
+    public enum ByteEnum : uint { Default, Byte }
```
am11: This one was supposed to be `: byte` instead of `: uint`, my bad.

YamlDotNet/Serialization/EventEmitters/JsonEventEmitter.cs
```diff
@@ -61,6 +61,18 @@ public override void Emit(ScalarEventInfo eventInfo, IEmitter emitter)
                 case TypeCode.UInt16:
                 case TypeCode.UInt32:
                 case TypeCode.UInt64:
+                    var valueIsEnum = eventInfo.Source.Value.GetType().IsEnum();
+                    if (valueIsEnum)
+                    {
+                        eventInfo.RenderedValue = eventInfo.Source.Value.ToString();
+                        eventInfo.Style = ScalarStyle.DoubleQuoted;
+                    }
+                    else
```
am11: nit: `else` case can be removed so control will flow to common path (avoid repetition):

```c#
...
case TypeCode.UInt64:
    if (eventInfo.Source.Value.GetType().IsEnum)
    {
        eventInfo.RenderedValue = eventInfo.Source.Value.ToString();
        eventInfo.Style = ScalarStyle.DoubleQuoted;
        break;
    }

case TypeCode.Single:
...
```

YamlDotNet/Serialization/EventEmitters/JsonEventEmitter.cs
```diff
@@ -61,6 +61,18 @@ public override void Emit(ScalarEventInfo eventInfo, IEmitter emitter)
                 case TypeCode.UInt16:
                 case TypeCode.UInt32:
                 case TypeCode.UInt64:
+                    var valueIsEnum = eventInfo.Source.Value.GetType().IsEnum();
```
aaubry: I think that this should use `eventInfo.Source.Type` instead of `eventInfo.Source.Value.GetType()`. The `Type` property already contains the resolved type, so there's no need to retrieve it again.

YamlDotNet/Serialization/EventEmitters/JsonEventEmitter.cs
KapuKapu: This does not compile. All execution path of a case must contain a brake or none at all.

aaubry:
Thanks!

Issue #398, opened 2019-05-03
Audit the failing tests from official YAML


aaubry:
As discussed in #395, audit (address or explain) the following failing tests from official YAML test suite:
- [x] [DK3J](https://github.com/yaml/yaml-test-suite/tree/data/DK3J)
- [x] [6M2F](https://github.com/yaml/yaml-test-suite/tree/data/6M2F)
- [x] [NJ66](https://github.com/yaml/yaml-test-suite/tree/data/NJ66)
- [x] [4MUZ](https://github.com/yaml/yaml-test-suite/tree/data/4MUZ)
- [x] [NHX8](https://github.com/yaml/yaml-test-suite/tree/data/NHX8)
- [x] [WZ62](https://github.com/yaml/yaml-test-suite/tree/data/WZ62)
- [x] [RTP8](https://github.com/yaml/yaml-test-suite/tree/data/RTP8)
- [x] [W5VH](https://github.com/yaml/yaml-test-suite/tree/data/W5VH)
- [x] [27NA](https://github.com/yaml/yaml-test-suite/tree/data/27NA)
- [x] [UDM2](https://github.com/yaml/yaml-test-suite/tree/data/UDM2)
- [x] [8XYN](https://github.com/yaml/yaml-test-suite/tree/data/8XYN)
- [x] [4ABK](https://github.com/yaml/yaml-test-suite/tree/data/4ABK)
- [x] [KZN9](https://github.com/yaml/yaml-test-suite/tree/data/KZN9)
- [x] [Q5MG](https://github.com/yaml/yaml-test-suite/tree/data/Q5MG)
- [x] [Y2GN](https://github.com/yaml/yaml-test-suite/tree/data/Y2GN)
- [x] [2JQS](https://github.com/yaml/yaml-test-suite/tree/data/2JQS)
- [x] [S3PD](https://github.com/yaml/yaml-test-suite/tree/data/S3PD)
- [x] [R4YG](https://github.com/yaml/yaml-test-suite/tree/data/R4YG)
- [x] [9SA2](https://github.com/yaml/yaml-test-suite/tree/data/9SA2)
- [x] [UT92](https://github.com/yaml/yaml-test-suite/tree/data/UT92)
- [x] [HWV9](https://github.com/yaml/yaml-test-suite/tree/data/HWV9)
- [x] [6ZKB](https://github.com/yaml/yaml-test-suite/tree/data/6ZKB)
[...]

PR #403, opened 2019-05-12
Relax anchor names allowed characters set

am11:
Except for the disallowed ones (`[`, `]`, `{`, `}`, `,`), all
characters are allowed (including unicode) in anchor name.

This delta fixes the remaining anchor spec tests. Two additional tests
are added to exercise corner case where anchor name has trailing colon
and alias is used in the key, and other with tag name containing emoji
characters.

* [W5VH](https://github.com/yaml/yaml-test-suite/tree/data/W5VH) (Allowed characters in alias)
* [8XYN](https://github.com/yaml/yaml-test-suite/tree/data/8XYN) (Anchor with unicode character)
* [Y2GN](https://github.com/yaml/yaml-test-suite/tree/data/Y2GN) (Anchor with colon in the middle)
* [2SXE](https://github.com/yaml/yaml-test-suite/tree/data/2SXE) (Anchors With Colon in Name)

Contributes to: #398


aaubry:
Thanks!

PR #405, opened 2019-05-17
Convert Samples to SDK style xunit project

am11:
Tests in Samples project were showing in VS' Test Explorer but on
execution, Output window was showing the following error:

> Test project YamlDotNet.Samples does not reference any .NET NuGet
> adapter. Test discovery or execution might not work for this project.
> It is recommended to reference NuGet test adapters in  each test
> project in the solution.

Adding the reference fixed the issue. Updated to new SDK-style format
for ease of editing (without unloading) and compact XML.


aaubry:
Thanks. I had just converted this and all other projects to SDK style, but the remaining changes were still helpful.

PR #406, opened 2019-05-20
 Improve omitted keys handling

### LINKED ISSUE #402: How to handle omitted key?
This is in relation to the PR https://github.com/aaubry/YamlDotNet/pull/401.

I started tackling the more comprehensive test: https://github.com/yaml/yaml-test-suite/blob/data/4ABK/test.event. It was failing line 6, 8 and 11 as below:

```diff
- master
+ PR

- =VAL ://foo.com
+ =VAL :http://foo.com

- =VAL :omitted value:
+ =VAL :omitted value

.. and line 11 throws on both
```

PR fixes first two issues. However for the third one,  `=VAL :omitted key`, scanner is classifying it as a `Value` token, which we ignore while generating events in parser and throw exception:

https://github.com/aaubry/YamlDotNet/blob/cc70e04bc88466a1432c9fbd5869c7efe49fbfad/YamlDotNet/Core/Parser.cs#L559-L562

I think there are at least two approaches:

* consider adding `string` value in `Value` token and case in parser `if(token is Value value) { /* generate scalar event here */}`
* update the scanner to classify omitted key as a (special) case of scalar

Which one would make more sense?

am11:

* [6M2F](https://github.com/yaml/yaml-test-suite/tree/data/6M2F) (Aliases in Explicit Block Mapping)
* [DBG4](https://github.com/yaml/yaml-test-suite/tree/data/DBG4) (Spec Example 7.10. Plain Characters)
* [4ABK](https://github.com/yaml/yaml-test-suite/tree/data/4ABK) (Spec Example 7.17. Flow Mapping Separate Values)
* [KZN9](https://github.com/yaml/yaml-test-suite/tree/data/KZN9) (Spec Example 7.21. Single Pair Implicit Entries)
* [FRK4](https://github.com/yaml/yaml-test-suite/tree/data/FRK4) (Spec Example 7.3. Completely Empty Flow Nodes)
* [9MMW](https://github.com/yaml/yaml-test-suite/tree/data/9MMW) (Spec Example 7.21. Single Pair Implicit Entries [1.3])

Contributes to: #398
Fixes #402

---

Additional two commits after rebasing with master were necessary for building clean repo with VS 2017.


YamlDotNet.Test/Spec/SpecTests.cs
```diff
@@ -39,12 +39,11 @@ public sealed class SpecTests
 
         private static readonly string specFixtureDirectory = GetTestFixtureDirectory();
 
-        // Note: all of these (29) tests are failing the assertion on line 65
```
am11: I have removed this line as it is not true anymore with this PR:
Spec test `S3PD` is failing at `Assert.Equal` (and missing the `VAL` node for omitted key). This is a corner case that both the key and value are omitted; single `:` on a separate line and the parser is expected to emit two empty `VAL` events. It would need further changes in scanner/parser. I have two more PRs in pipe, then I will attempt to address this case.

aaubry:
I was unable to merge this sooner. Sorry

am11:
No problems. :)

PR #416, opened 2019-06-25
Allow empty value for non-specific tag

am11:

* [52DL](https://github.com/yaml/yaml-test-suite/tree/data/52DL) (Explicit Non-Specific Tag [1.3])
* [S4JQ](https://github.com/yaml/yaml-test-suite/tree/data/S4JQ) (Spec Example 6.28. Non-Specific Tags)
* [8MK2](https://github.com/yaml/yaml-test-suite/tree/data/8MK2) (Explicit Non-Specific Tag)

Contributes to: #398


aaubry:
Thanks! Sorry for the delay.

PR #418, opened 2019-07-08
Add ability to implicitly start a bare document

am11:

* [M7A3](https://github.com/yaml/yaml-test-suite/tree/data/M7A3) (Spec Example 9.3. Bare Documents)

Contributes to: #398


aaubry:
Thanks!

PR #420, opened 2019-07-17
Ignore reserved directives for future use

am11:

* [2LFX](https://github.com/yaml/yaml-test-suite/tree/data/2LFX) (Spec Example 6.13. Reserved Directives [1.3])
* [6LVF](https://github.com/yaml/yaml-test-suite/tree/data/6LVF) (Spec Example 6.13. Reserved Directives)

Contributes to: #398

---

In a separate commit, I have added license file to nupkg to fix NU5125 deprecation warning that shows up in AppVeyor.
Related refs:
* https://github.com/NuGet/docs.microsoft.com-nuget/blob/master/docs/reference/nuspec.md#license
* https://github.com/NuGet/docs.microsoft.com-nuget/blob/master/docs/reference/msbuild-targets.md#pack-target
* https://github.com/NuGet/docs.microsoft.com-nuget/blob/master/docs/reference/msbuild-targets.md#packing-a-license-expression-or-a-license-file
* https://github.com/NuGet/Home/issues/7601 (the reason why the file is renamed to LICENSE.txt, with extension)


aaubry:
Nice. Thanks!

PR #421, opened 2019-07-18
Respect DocumentEnd while scanning the block scalar

am11:
Also extracted `IsDocumentStart` and `IsDocumentEnd` as <del>readonly
properties</del> separate methods.
<sup>(changed from prop to method for alignment with existing code)</sup>

* [W4TN](https://github.com/yaml/yaml-test-suite/tree/data/W4TN) (Spec Example 9.5. Directives Documents)

Contributes to: #398


aaubry:
Thanks for all this work! I really appreciate it.

PR #422, opened 2019-07-19
Test the exactness of parser output in error cases

am11:
This delta adds another ignore list for error cases, where YDN's
parser/scanner correctly raises the exception but the parser output
(till the point of error) mismatches what is expected by the spec test.
In some cases we scan more tokens and in others less.

Fixing these would probably mean being pedantic, but that would also
harden the conformance status.

Related to: #398

### DIFF
--- a/YamlDotNet.Test/Spec/SpecTests.cs
+++ b/YamlDotNet.Test/Spec/SpecTests.cs
@@ -49,6 +49,12 @@ namespace YamlDotNet.Test.Spec
             "X4QW", "9C9N", "QB6E", "CVW2", "9JBA", "HRE5", "SU5Z", "QLJ7"
         };
 
+        private static readonly List<string> knownParserDesyncInErrorCases = new List<string>
+        {
+            "2CMS", "3HFZ", "4EJS", "4H7K", "4JVG", "7MNF", "9CWY", "9KBC", "BS4K", "C2SP",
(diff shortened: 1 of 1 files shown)


aaubry:
I see the problem. It's hard to tell whether scanning too many or too less tokens is relevant to the test, or if it is an implementation detail. Maybe we should review each tests individually ?

am11:
In some of these failing cases where we read more than expected tokens, the cause of error YDN has is different than what spec is expecting. This is somewhat failure by accident that the error test case leads to some other subsequent error, which YDN correctly catches and raise exception for. In that regard, fixing those cases would still be matter of correctness.

> or if it is an implementation detail

In other cases where we read less tokens, it is due to the implementation detail that when the parser moves the scanner for the next token, scanner reads one or many tokens at once (in batch). If one of the token parsing throws, the whole transaction is concluded at that point and parsing stops. For this, I have added a new `internal` token type `Error`, that scanner uses instead of exception (in cases where spec test fail). This way parser respects scanner's best effort and emits events for all the scanned tokens. When parser encounters the token type `Error`, it throws the exception with the message value from the token (`While scanning the <token kind>, ..`).

This is bit pedantic, but it also ensures that we are erroring out for the same reason as the spec is expecting. Over at Sass, we ended up standardizing the error messages as well, to prevent implementer from second guessing (e.g. https://github.com/sass/sass-spec/blob/0688208/spec/errors/invalid-operation/mod.hrx).

I am going through each case and fixing the parser/scanner (at 11/20 now), once the full change is ready, I will split it into separate commits and send a pull request for review.

PR #435, opened 2019-09-28
Change the default value handling behavior

### LINKED ISSUE #427: Change the default serializer behaviour to emit default values
As evidenced by #318, #310, #298, #268, #251, #258, #153, the current behaviour of omitting default values during serialization is misleading and a source of confusion.

I'm opening this issue so that we can discuss the best way to handle this. I suggest the following changes, but feel free to propose alternatives:
- (1) **Change the default behaviour to emit *all* default values**, including null references. This will be a breaking change but it is necessary to move forward.
- (2) **Add an option to not emit any default value**. This ensures a simple upgrade path for anyone who depends on the current behaviour.

Additionally, we could offer some or all of the following options:
- (3) **Add an option to not emit any null value**. This was the original intention and I still believe that it is a valid use case.
- (4) **Add a property to `YamlMemberAttribute`** to override the emission of defaults on a specific member.
- (5) **Add an option to not emit default values for a specific type**. Would this be useful ?

I think that performing changes (1) and (2) is necessary to properly address the issue. I am inclined on also including (3) and (4) because they offer more flexibility. I don't see any use case for (5), but if someone can come up with a relevant one, let's discuss it.

For more complex behaviour, it is already possible to replace `DefaultExclusiveObjectGraphVisitor` by a custom implementation that can decide whether to emit any value based on arbitrary rules.

aaubry:
Fixes #427

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,10 @@ install:
   #
   - if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
       git checkout $TRAVIS_BRANCH;
+      git reset $TRAVIS_COMMIT --hard;
     else
       git checkout -b $TRAVIS_PULL_REQUEST_BRANCH > /dev/null;
     fi
-  - git reset $TRAVIS_COMMIT --hard
-  - git status
   - docker pull aaubry/yamldotnet
 
 script:
--- a/YamlDotNet.Test/Serialization/EmitDefaultValuesTests.cs
+++ b/YamlDotNet.Test/Serialization/EmitDefaultValuesTests.cs
@@ -1,0 +1,131 @@
+﻿using System.ComponentModel;
+using Xunit;
+using YamlDotNet.Serialization;
+
+namespace YamlDotNet.Test.Serialization
+{
+    public class EmitDefaultValuesTests
+    {
+        private class Model
+        {
+            public string ANullString => null;
+            [DefaultValue("hello")] public string ADefaultString => "hello";
+            [DefaultValue("hello")] public string ANonDefaultString => "world";
+            [DefaultValue("hello")] public string ANonDefaultNullString => null;
+
+            public int AZeroInteger => 0;
+            public int ANonZeroInteger => 1;
+            [DefaultValue(2)] public int ADefaultInteger => 2;
+            [DefaultValue(2)] public int ANonDefaultZeroInteger => 0;
+
+            public int? ANullInteger => null;
+            public int? ANullableZeroInteger => 0;
+            public int? ANullableNonZeroInteger => 1;
+            [DefaultValue(2)] public int? ANullableNonZeroDefaultInteger => 2;
+            [DefaultValue(2)] public int? ANullableNonZeroNonDefaultInteger => 1;
+        }
+
+        [Fact]
+        public void All_default_values_and_nulls_are_emitted_when_no_configuration_is_performed()
+        {
+            // Arrange
+            var sut = new SerializerBuilder()
+                .Build();
+
+            // Act
+            var yaml = sut.Serialize(new Model());
+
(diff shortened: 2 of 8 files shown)


aaubry:
This changes the default behaviour of the serializer to always emit null and default values. This behavior can also be controlled using `ConfigureDefaultValuesHandling` on the SerializerBuilder, or through the YamlMember attribute. 

The documentation has been updated in a branch, but since GitHub cannot render wiki from branches other than `master`, I'll reproduce the change here. This is the part of the documentation of `SerializerBuilder` that describes the new `ConfigureDefaultValuesHandling` method, that replaces `EmitDefaults`.

### `ConfigureDefaultValuesHandling(DefaultValuesHandling)`

The default behaviour of the serializer is to emit all properties regardless of their value. The behaviour can be modified if one prefers not to see null or default values. The `configuration` parameter is used to specify how properties with null or default values should be handled:

| Value | Behavior |
|-|-|
| Preserve | Specifies that all properties are to be emitted regardless of their value. This is the default behavior. |
| OmitNull | Specifies that properties that contain null references or a null `Nullable<T>` are to be omitted. |
| OmitDefaults | Specifies that properties that that contain their default value, either `default(T)` or the value specified in [`DefaultValueAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.defaultvalueattribute) are to be omitted. |

The following example illustrates the difference between the behaviours. Consider the following class:

```c# id=nullsanddefaults
class Model
{
    public string NullString => null;
    [DefaultValue("some-default")] public string NullStringWithDefault => null;
    public int ZeroInteger => 0;
    public int? NullableInteger => null;
    [DefaultValue(42)] public int DefaultInteger => 42;
}
```

Compare the outputs depending on the `DefaultValuesHandling` configuration:

```c# include=nullsanddefaults
var model = new Model();

Console.WriteLine("# DefaultValuesHandling.Preserve (default)");
new SerializerBuilder()
    .ConfigureDefaultValuesHandling(DefaultValuesHandling.Preserve)
    .Build()
    .Serialize(Console.Out, model);

Console.WriteLine("# DefaultValuesHandling.OmitNull");
new SerializerBuilder()
    .ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitNull)
    .Build()
    .Serialize(Console.Out, model);

Console.WriteLine("# DefaultValuesHandling.OmitDefaults");
new SerializerBuilder()
    .ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitDefaults)
    .Build()
    .Serialize(Console.Out, model);
```

```yaml output
# DefaultValuesHandling.Preserve (default)
NullString: 
NullStringWithDefault: 
ZeroInteger: 0
NullableInteger: 
DefaultInteger: 42
# DefaultValuesHandling.OmitNull
ZeroInteger: 0
DefaultInteger: 42
# DefaultValuesHandling.OmitDefaults
NullStringWithDefault: 
```

aaubry:
I have published a pre-release package named `8.0.0-emit-default-values0565` if anyone wants to try this.

MiYanni:
I tested the `8.0.0-emit-default-values0565` package today as I was running into the enum serialization issue (not serializing the 0th value). I used `.ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitNull)` since I do not need null values to be serialized. The enums are now being serialized as expected! 😄

aaubry:
I'm glad that this change works for you, @MiYanni. Did you feel the need for more control, or are the current configuration options enough for your use case?

MiYanni:
@aaubry Sorry for not responding sooner. I did not find I needed finer control for my use case. Thanks for implementing this change! 👍

aaubry:
No worries. Thanks for the feedback!

Issue #450, opened 2019-11-08
Serializer emits default values


greuelpirat:
After updating to 8.0.0 the Serializer

```
var serializer = serializerBuilder
                    .DisableAliases()
                    .BuildValueSerializer();
```

is emitting default values altough .EmitDefault() is not called. Has this behaviour changed? How can I disable this with 8.0.0?

aaubry:
Yes, this behaviour has changed. You can read more about it in the [release notes](https://github.com/aaubry/YamlDotNet/releases/tag/v8.0.0), which you should always read before upgrading to a different major version.

greuelpirat:
I must apologize. I did not check the release notes. sorry.

aaubry:
No worries 👍. I realize that my response was a bit harsh, sorry about that.

PR #457, opened 2019-12-08
Added async to serialize function, and created async unit tests to verify

samsmithnz:
TODO, before merging Pull Request:
- [x] Update conditional statements to support all .NET Standard, .NET Core, and back to .NET Framework 4.5 
```#if` NETSTANDARD && !NETSTANDARD1_3```
- [x] Create more tests for serialization
- [ ] Resolve performance build issues
- [ ] Add async around deserialization (Not sure there is much async to be done here)

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -1289,7 +1289,7 @@ namespace YamlDotNet.Test.Serialization
                 .Build();
 
             var writer = new StringWriter();
-#if NETSTandard && !NETSTANDARD1_3
+#if NETSTANDARD && !NETSTANDARD1_3
             await serializer.SerializeAsync(writer, obj);
 #else
             serializer.Serialize(writer, obj);
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -1253,6 +1253,7 @@ namespace YamlDotNet.Test.Serialization
             Assert.Equal(3, exception.Start.Column);
         }
 
+        [Trait("Category","Async")]
         [Fact]
         public void SerializeDynamicPropertyAndApplyNamingConvention()
         {
@@ -1273,7 +1274,7 @@ namespace YamlDotNet.Test.Serialization
             writer.ToString().Should().Contain("xxx: new_value");
         }
 
-
+        [Trait("Category","Async")]
         [Fact]
         public async System.Threading.Tasks.Task SerializeDynamicPropertyAndApplyNamingConventionAsync()
         {
@@ -1548,6 +1549,7 @@ namespace YamlDotNet.Test.Serialization
             Assert.NotNull(deserialized);
         }
 
+        [Trait("Category","Async")]
         [Fact]
         public void SerializeExceptionWithStackTrace()
         {
@@ -1559,6 +1561,7 @@ namespace YamlDotNet.Test.Serialization
             Assert.Contains("GetExceptionWithStackTrace", yaml);
         }
 
+        [Trait("Category","Async")]
         [Fact]
         public async System.Threading.Tasks.Task SerializeExceptionWithStackTraceAsync()
         {
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -1276,7 +1276,9 @@ namespace YamlDotNet.Test.Serialization
 
         [Trait("Category","Async")]
         [Fact]
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         public async System.Threading.Tasks.Task SerializeDynamicPropertyAndApplyNamingConventionAsync()
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
         {
             dynamic obj = new ExpandoObject();
             obj.property_one = new ExpandoObject();
@@ -1563,7 +1565,9 @@ namespace YamlDotNet.Test.Serialization
 
         [Trait("Category","Async")]
         [Fact]
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         public async System.Threading.Tasks.Task SerializeExceptionWithStackTraceAsync()
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
         {
             var ex = GetExceptionWithStackTrace();
             var serializer = new SerializerBuilder()
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -45,7 +45,7 @@ namespace YamlDotNet.Test.Serialization
 {
     public class SerializationTests : SerializationTestHelper
     {
-        #region Test Cases
+#region Test Cases
 
         private static readonly string[] TrueStrings = { "true", "y", "yes", "on" };
         private static readonly string[] FalseStrings = { "false", "n", "no", "off" };
@@ -68,7 +68,7 @@ namespace YamlDotNet.Test.Serialization
             }
         }
 
-        #endregion
+#endregion
 
         [Fact]
         public void DeserializeEmptyDocument()
@@ -1112,7 +1112,7 @@ namespace YamlDotNet.Test.Serialization
                   - &BIG { r: 10 }
                   - &SMALL { r: 1 }
                 
-                # All the following maps are equal:
+# All the following maps are equal:
                 results:
                   - # Explicit keys
                     x: 1
@@ -1130,12 +1130,12 @@ namespace YamlDotNet.Test.Serialization
                     label: center/big
                   
                   - # Override
-                    #<< : [ *BIG, *LEFT, *SMALL ]    # This does not work because, in the current implementation,
-                                                     # later keys override former keys. This could be fixed, but that
-                                                     # is not trivial because the deserializer allows aliases to refer to
-                                                     # an anchor that is defined later in the document, and the way it is
-                                                     # implemented, the value is assigned later when the anchored value is
-                                                     # deserialized.
+#<< : [ *BIG, *LEFT, *SMALL ]    # This does not work because, in the current implementation,
+# later keys override former keys. This could be fixed, but that
+# is not trivial because the deserializer allows aliases to refer to
+# an anchor that is defined later in the document, and the way it is
+# implemented, the value is assigned later when the anchored value is
+# deserialized.
                     << : [ *SMALL, *LEFT, *BIG ]
                     x: 1
                     label: center/big
@@ -1556,6 +1556,22 @@ namespace YamlDotNet.Test.Serialization
                 .WithTypeConverter(new MethodInfoConverter())
                 .Build();
             string yaml = serializer.Serialize(ex);
+            Assert.Contains("GetExceptionWithStackTrace", yaml);
+        }
+
+        [Fact]
+        public async System.Threading.Tasks.Task SerializeExceptionWithStackTraceAsync()
+        {
+            var ex = GetExceptionWithStackTrace();
+            var serializer = new SerializerBuilder()
+                .WithTypeConverter(new MethodInfoConverter())
+                .Build();
+
+#if NETSTANDARD && !NETSTANDARD1_3
+           string yaml = await serializer.SerializeAsync(ex);
+#else
+            string yaml = serializer.Serialize(ex);
+#endif
             Assert.Contains("GetExceptionWithStackTrace", yaml);
         }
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by samsmithnz** [comment]
Odd, seem to be receiving some errors from the performance tests that I haven't touched yet: 

```
ReceiptTest.cs(24,35): error CS0234: The type or namespace name 'Lib' does not exist in the namespace 'YamlDotNet.PerformanceTests' (are you missing an assembly reference?) [/build/PerformanceTests/YamlDotNet.PerformanceTests.v8.0.0/YamlDotNet.PerformanceTests.v8.0.0.csproj]
ReceiptTest.cs(33,26): error CS0246: The type or namespace name 'Receipt' could not be found (are you missing a using directive or an assembly reference?) [/build/PerformanceTests/YamlDotNet.PerformanceTests.v8.0.0/YamlDotNet.PerformanceTests.v8.0.0.csproj]
Build FAILED.
ReceiptTest.cs(24,35): error CS0234: The type or namespace name 'Lib' does not exist in the namespace 'YamlDotNet.PerformanceTests' (are you missing an assembly reference?) [/build/PerformanceTests/YamlDotNet.PerformanceTests.v8.0.0/YamlDotNet.PerformanceTests.v8.0.0.csproj]
ReceiptTest.cs(33,26): error CS0246: The type or namespace name 'Receipt' could not be found (are you missing a using directive or an assembly reference?) [/build/PerformanceTests/YamlDotNet.PerformanceTests.v8.0.0/YamlDotNet.PerformanceTests.v8.0.0.csproj]
    0 Warning(s)
    2 Error(s)
```

File: YamlDotNet.Test/Serialization/SerializationTests.cs [hunk 1]
```diff
@@ -1273,6 +1273,32 @@ public void SerializeDynamicPropertyAndApplyNamingConvention()
             writer.ToString().Should().Contain("xxx: new_value");
         }
 
+
+        [Fact]
```
samsmithnz: Note that if I put the pre-compiler condition around the entire function, it doesn't show in the test explorer and doesn't seem to run. This is odd behavior I'm unable to resolve. This implementation produces two warnings that I can ignore with some overrides later if this is the "best" solution

YamlDotNet.Test/YamlDotNet.Test.csproj
```diff
@@ -39,4 +39,9 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+
+  <!--Add Async for projects that are not netstandard 2.0 or higher (due to System.Linq.Async dependency)-->
+  <ItemGroup Condition="'$(NetStandard)' == 'true' And '$(TargetFramework)' != 'netstandard1.3'">
```
samsmithnz: This will technically support all of .NET Standard and .NET Framework 4.5, which I will fix later, this was trying to initially target .NET Standard 2+ and/or .NET Core 3.

YamlDotNet.sln
```diff
@@ -52,7 +52,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docker", "docker", "{45384A
 		docker\Dummy.csproj = docker\Dummy.csproj
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "YamlDotNet.PerformanceTests.v8.0.0", "PerformanceTests\YamlDotNet.PerformanceTests.v8.0.0\YamlDotNet.PerformanceTests.v8.0.0.csproj", "{C0363835-93DB-4AF1-9E3C-0689C65D0FD7}"
```
samsmithnz: Unclear why this changed - perhaps it's related to the errors in the performance testing I'm seeing in the travis build...

samsmithnz:
Some current unit test results to date: 3 functions see some performance degradation with async, but overall, there is a big performance gain: image

aaubry:
This looks promising! I don't have much today but I will try to have a look tomorrow.

YamlDotNet.Test/YamlDotNet.Test.csproj
samsmithnz: updated to <ItemGroup Condition="'$(NetStandard)' == 'true' Or '$(TargetFramework)' == 'net45'">

YamlDotNet/Serialization/ISerializer.cs
```diff
@@ -13,12 +16,29 @@ public interface ISerializer
...
         /// </summary>
         /// <param name="graph">The object to serialize.</param>
         string Serialize(object graph);
 
+#if NETSTANDARD || NET45
+        /// <summary>
+        /// Serializes the specified object into a string.
+        /// </summary>
+        /// <param name="graph">The object to serialize.</param>
+        Task<string> SerializeAsync(object graph);
```
aaubry: There's probably no point in having an async version of this method since it works in-memory.

YamlDotNet.Test/Serialization/SerializationTests.cs
```diff
@@ -1112,7 +1199,7 @@ public void ExampleFromSpecificationIsHandledCorrectly()
                   - &BIG { r: 10 }
                   - &SMALL { r: 1 }
                 
-                # All the following maps are equal:
+# All the following maps are equal:
```
aaubry: This indentation should be kept as it improves the readability of the document. I suppose that it was removed by accident.

YamlDotNet/Serialization/LazyComponentRegistrationListExtensions.cs
```diff
@@ -54,11 +57,29 @@ public static List<TComponent> BuildComponentList<TComponent>(this LazyComponent
                 .ToList();
         }
 
+#if NETSTANDARD || NET45
+        public static async Task<List<TComponent>> BuildComponentListAsync<TComponent>(this LazyComponentRegistrationList<Nothing, TComponent> registrations)
```
aaubry: Why is this method async ?

YamlDotNet/Serialization/LazyComponentRegistrationListExtensions.cs
```diff
@@ -54,11 +57,29 @@ public static List<TComponent> BuildComponentList<TComponent>(this LazyComponent
...
+
         public static List<TComponent> BuildComponentList<TArgument, TComponent>(this LazyComponentRegistrationList<TArgument, TComponent> registrations, TArgument argument)
         {
             return registrations
                 .Select(factory => factory(argument))
                 .ToList();
         }
+
+#if NETSTANDARD || NET45
+        public static async Task<List<TComponent>> BuildComponentListAsync<TArgument, TComponent>(this LazyComponentRegistrationList<TArgument, TComponent> registrations, TArgument argument)
```
aaubry: Why is this method async ?

YamlDotNet/Serialization/SerializerBuilder.cs
```diff
@@ -566,6 +569,31 @@ public void SerializeValue(IEmitter emitter, object? value, Type? type)
...
+                }
+
+                void nestedObjectSerializer(object? v, Type? t) => SerializeValue(emitter, v, t);
+
+                var emittingVisitor = emissionPhaseObjectGraphVisitorFactories.BuildComponentChain(
+                    new EmittingObjectGraphVisitor(eventEmitter),
+                    inner => new EmissionPhaseObjectGraphVisitorArgs(inner, eventEmitter, preProcessingPhaseObjectGraphVisitors, typeConverters, nestedObjectSerializer)
+                );
+
+                traversalStrategy.Traverse(graph, emittingVisitor, emitter);
```
aaubry: This is the part that should be async

YamlDotNet/Serialization/SerializerBuilder.cs
```diff
@@ -566,6 +569,31 @@ public void SerializeValue(IEmitter emitter, object? value, Type? type)
...
+
+#if NETSTANDARD || NET45
+            public async Task SerializeValueAsync(IEmitter emitter, object? value, Type? type)
+            {
+                var actualType = type ?? (value != null ? value.GetType() : typeof(object));
+                var staticType = type ?? typeof(object);
+
+                var graph = new ObjectDescriptor(value, actualType, staticType);
+
+                var preProcessingPhaseObjectGraphVisitors = await preProcessingPhaseObjectGraphVisitorFactories.BuildComponentListAsync(typeConverters);
```
aaubry: There's nothing to perform asynchronously here. As mentioned earlier, there's no need for an async variant of BuildComponentList

YamlDotNet.Test/Serialization/SerializationTests.cs
```diff
@@ -1130,12 +1217,12 @@ public void ExampleFromSpecificationIsHandledCorrectly()
...
-                                                     # is not trivial because the deserializer allows aliases to refer to
-                                                     # an anchor that is defined later in the document, and the way it is
-                                                     # implemented, the value is assigned later when the anchored value is
-                                                     # deserialized.
+#<< : [ *BIG, *LEFT, *SMALL ]    # This does not work because, in the current implementation,
+# later keys override former keys. This could be fixed, but that
+# is not trivial because the deserializer allows aliases to refer to
+# an anchor that is defined later in the document, and the way it is
+# implemented, the value is assigned later when the anchored value is
+# deserialized.
```
aaubry: This indentation should be kept as it improves the readability of the document. I suppose that it was removed by accident.

YamlDotNet/Serialization/Serializer.cs
```diff
@@ -142,5 +225,19 @@ private void EmitDocument(IEmitter emitter, object graph, Type? type)
             emitter.Emit(new DocumentEnd(true));
             emitter.Emit(new StreamEnd());
         }
+
+#if NETSTANDARD || NET45
+        private async Task EmitDocumentAsync(IEmitter emitter, object graph, Type? type)
+        {
+            emitter.Emit(new StreamStart());
+            emitter.Emit(new DocumentStart());
```
aaubry: This should use EmitAsync the emitter will perform IO

YamlDotNet/Serialization/Serializer.cs
```diff
@@ -142,5 +225,19 @@ private void EmitDocument(IEmitter emitter, object graph, Type? type)
...
+#if NETSTANDARD || NET45
+        private async Task EmitDocumentAsync(IEmitter emitter, object graph, Type? type)
+        {
+            emitter.Emit(new StreamStart());
+            emitter.Emit(new DocumentStart());
+
+            await valueSerializer.SerializeValueAsync(emitter, graph, type);
+
+            emitter.Emit(new DocumentEnd(true));
+            emitter.Emit(new StreamEnd());
```
aaubry: This should use EmitAsync the emitter will perform IO

samsmithnz:
> I've reviewed your changes so far, and added a few comments. Regarding the tests, there should probably be tests that perform actual IO, since reading or writing from memory will always be a synchronous operation.

@aaubry Tons of interesting comments here, thanks for this. Before I dive into the details, I'm wondering if there is confusion about the goal of this PR. Are you confusing multi-threading with **async** support? My goal was to add async support: 

> _Async methods are intended to be non-blocking operations. An await expression in an async method doesn’t block the current thread while the awaited task is running. Instead, the expression signs up the rest of the method as a continuation and returns control to the caller of the async method._

> _**The async and await keywords don't cause additional threads to be created.** Async methods don't require multithreading because an async method doesn't run on its own thread. The method runs on the current synchronization context and uses time on the thread only when the method is active._ 

Reading and writing from memory is IO - so there is a benefit to using async here - although you are correct that it's not as much of a benefit as writing to a file. I started with serialize as an example, and I don't see files being testing directly here, so I don't have tests here yet either.

YamlDotNet/Serialization/LazyComponentRegistrationListExtensions.cs
samsmithnz: Trying to use ToListAsync, but it's been replaced: https://stackoverflow.com/a/57851481/337421

YamlDotNet.Test/Serialization/SerializationTests.cs
samsmithnz: My mistake, I instinctively did auto format on the class and accidentally changed this (Ctrl+K, Ctrl+F). I will revert.

aaubry:
@samsmithnz I'm not sure whether there is confusion or not. What I am saying is that memory IO is non-blocking so there's no benefit in using async APIs to access it.
Regarding builder methods such as `BuildComponentList`, I don't see a point in them being async as they are just configuring the serializer. They are not performing any IO, so async does not apply.

Maybe it would be easier to start adding async methods to the `Emitter` and then build up from there, instead of starting with the serializer.

samsmithnz:
I just saw the note in #450. I'm unlikely to continue with this anytime soon. Please consider it abandoned. Sorry I couldn't deliver.

samsmithnz:
Closing/abandoning PR.

PR #460, opened 2019-12-21
Fix type conflicts when targeting frameworks without nullability attributes

### LINKED ISSUE #456: YamlDotNet shouldn't export NRT attributes.
YamlDotNet defines [`NotNullWhenAttribute` and `MaybeNullWhenAttribute`](https://github.com/aaubry/YamlDotNet/blob/b6cf0f6c9de00b30499607b4bade362f72f2f21f/YamlDotNet/Helpers/Portability.cs#L751) as public.

These really should be internal so that they don't introduce name conflicts when this package is consumed in conjunction with another that defines them.

The issue can be worked around with an alias, but shouldn't have to.

Nihlus:
In its current form, YamlDotNet produces type conflicts with other projects that target frameworks which do not include nullability attributes, but add them in other ways to allow full nullability support in the compiler. The issue typically displays as a warning during compilation that one or the other type was selected.

```
Appearance.cs(255, 14): [CS0436] The type 'NotNullWhenAttribute' in '/[PROJECT]/obj/Debug/netstandard2.0/NuGet/16700ED22EA87F787941E08E0F8DABDD3F92ADDC/Nullable/1.1.1/Nullable/NullableAttributes.cs' conflicts with the imported type 'NotNullWhenAttribute' in 'YamlDotNet, Version=8.0.0.0, Culture=neutral, PublicKeyToken=ec19458f3c15af5e'. Using the type defined in '[PROJECT]/obj/Debug/netstandard2.0/NuGet/16700ED22EA87F787941E08E0F8DABDD3F92ADDC/Nullable/1.1.1/Nullable/NullableAttributes.cs'.
```

The issue is caused by YamlDotNet declaring the polyfill attributes in Portability.cs as `public`, and not covering all frameworks in the preprocessor directives that guard them.

This pull request removes YamlDotNet's own polyfill attributes, and references a compile-time NuGet packages which defines all nullability attributes and compiles them as internal attributes into targets which require them, preventing any conflicts with other projects. The package used ([Nullable](https://github.com/manuelroemer/Nullable)) is completely transparent to the end user, and does not incur any additional dependencies for consumers.

Fixes #456.

### DIFF
--- a/YamlDotNet/Helpers/Portability.cs
+++ b/YamlDotNet/Helpers/Portability.cs
@@ -244,7 +244,7 @@ namespace YamlDotNet
         {
             return ex.InnerException;
         }
-        
+
         public static bool IsInstanceOf(this Type type, object o)
         {
             return o.GetType() == type || o.GetType().GetTypeInfo().IsSubclassOf(type);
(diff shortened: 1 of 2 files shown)


Nihlus:
I'm unsure why the build failed. It does not appear to have anything to do with my changes.

aaubry:
Thanks for this. I will do a release today that includes this.

Nihlus:
Great!

aaubry:
YamlDotNet 8.1.0 has just been uploaded to NuGet.org. It should be available in a few minutes.

PR #464, opened 2020-02-05
Update ScalarNodeDeserializer.cs

ETSD:
Removed support for Octal numbers as this is no longer supported in the YAML specification.

### DIFF
--- a/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
+++ b/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
@@ -149,40 +149,32 @@ namespace YamlDotNet.Serialization.NodeDeserializers
             if (value[currentIndex] == '0')
             {
                 // Could be binary, octal, hex, decimal (0)
+                
+                // Check the next character
+                currentIndex++;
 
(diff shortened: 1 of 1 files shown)


aaubry:
What's the point of removing this support ? YamlDotNet implements the YAML 1.1 specification, which specifies octal numbers. Even the 1.2 specification mentions octal numbers, although the format is different.

I don't know if anyone is relying on octal numbers but we can't just remove that support.
I could accept a way to disable octal support, although ideally that would be driven by the version of the YAML document.

ETSD:
We have some coding guidelines that we are required to adhere to. Included in these is that all numbers be prefixed with '0' characters (usually to three or four characters). Multiple developers have run into this "feature" from an outdated YAML specification. Given that there is no way within C# to explicitly identify a number as Octal (unlike Binary and Hex), and that this is a library specifically for C#, it is an odd choice to assume that numbers prefixed with a '0' character (if not explicitly identified as Binary or Hex) should be treated as Octal.

aaubry:
I understand your need, but this needs to be configurable. I'd like to avoid introducing breaking changes unless they are inevitable.

EdwardCooke:

EdwardCooke:

PR #468, opened 2020-02-17
Get the yaml mapping node to preserve the order of the keys

a2937:
For this pull request, I am using a wrapper of the OrderedDictionary class to hold and manage all of the data stored in the YamlMappingNode. This wrapper was implemented because the OrderedDictionary class was unfortunately not generic , and I could not find a Generic version of the class in the System.Collections.Generic namespace.


aaubry:
I've checked your implementation and noticed that many of the IDictionary members are not implemented. This is a problem because users of the code may rely on those members. Maybe it would be simpler to use both a `Dictionary<YamlNode, YamlNode>` and a `List<YamlNode>` internally. This would also have the advantage of avoiding taking a new dependency on System.Collections.Specialized.
What do you think ?

a2937:
I implemented more of the IDictionary functions and it actually resolved the serialization issues. Funny how that works. The only downside is that the library will not be supported by Dotnet Standard 1.3. I've been trying to find another collection to use to serialize with instead; however I have found no such luck. I am still insisting on using the OrderedDictionary class. It is the only class that I am assured will keep the order of the key value insertions.

aaubry:
I'm sorry, I was unable to get back to you earlier. I'd rather avoid System.Collections.Specialized. This forces that dependency to anyone using the library while providing limited value in this case. If you look at the implementation of `OrderedDictionary`, you'll see that it consists of a dictionary and a list. Methods that add or remove add to both collections and enumeration iterates the list.
If you use this approach you will also benefit from generics which will avoid any casts.

Regarding serialization, that's not a use case that we intend to keep supporting. These tests are obsolete and I intended to remove them earlier but forgot about it. You may delete BinarySerlializationTests.cs and ignore serialization.

a2937:
Okay I swapped to a Dictionary and List but can't quite figure out how to keep them from being loaded back as YamlNodes instead of YamlAliasNodes. Is it possible that I removed a library I shouldn't have or am using something incorrectly?

aaubry:
I am confused about your last question. Can you explain ?

a2937:
I'm suddenly getting this error and don't know how to fix it or why it is happening : "System.NotSupportedException : A YamlAliasNode is an implementation detail and should never be visited."

airbreather:
I was about to open an issue requesting this, but I see that there's already this PR for it, so I'll just comment here.

+1 on strengthening the `YamlMappingNode` contract to preserve the order of keys; the workaround I've found is to use a sequence of 1-element mapping nodes, which is clunky and doesn't automatically enforce uniqueness.

I've looked at this, and while there are a few things I'd do slightly differently here and there, the only thing I'm going to suggest at this time (as an outsider) is to change `OrderedYamlDictionary` accessibility from `public` to `internal`.  With `public`, people might start using it for their own needs, which could complicate maintenance efforts in the future.

airbreather:
Ahh, was this actually done in #542?

EdwardCooke:
This was already implemented in another PR, closing.

PR #469, opened 2020-02-17
Fix few parser desync cases

am11:

* [4H7K](https://github.com/yaml/yaml-test-suite/tree/data/4H7K) (Flow sequence with invalid extra closing bracket)
* [4JVG](https://github.com/yaml/yaml-test-suite/tree/data/4JVG) (Scalar value with two anchors)
* [C2SP](https://github.com/yaml/yaml-test-suite/tree/data/C2SP) (Flow Mapping Key on two lines)
* [EB22](https://github.com/yaml/yaml-test-suite/tree/data/EB22) (Missing document-end marker before directive)

Contributes to: #398

### DIFF
--- a/YamlDotNet.Test/Spec/SpecTests.cs
+++ b/YamlDotNet.Test/Spec/SpecTests.cs
@@ -51,7 +51,7 @@ namespace YamlDotNet.Test.Spec
 
         private static readonly List<string> knownParserDesyncInErrorCases = new List<string>
         {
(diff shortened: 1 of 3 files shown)


am11:
Found some time to tackle these four. The remaining two are quite pedantic and would require some rationalization by rereading potions of spec doc. :)

aaubry:
Nice work!

PR #470, opened 2020-02-18
Switch to C# classlib GUIDs for all projects in SLN

am11:
Whenever we add/remove/re-add project to sln using VS (as opposed to [`dotnet-sln` commands](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-sln#commands)), VS replaces all existing guids with `9A19103F-16F7-4668-BE54-9A1E7A4F7556`, as happened in https://github.com/aaubry/YamlDotNet/commit/e6c355dabd6e2a36ae8df8f1cab8b8c50dbb1b02.

My VS was having hard time discovering tests, and similar problems as seen in: https://github.com/aaubry/YamlDotNet/pull/405.

Until Microsoft fixes this VS bug https://github.com/dotnet/project-system/issues/1821, we would need to manually undo the automagical replacements of this guid. Also added validation in `build.sh` to fail the build in the CI machine, if it gets accidentally checked in.

### DIFF
--- a/YamlDotNet.sln
+++ b/YamlDotNet.sln
@@ -3,39 +3,43 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.29123.88
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "YamlDotNet", "YamlDotNet\YamlDotNet.csproj", "{ABFC2E37-119B-439D-82B6-49E9680EEA90}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "YamlDotNet", "YamlDotNet\YamlDotNet.csproj", "{ABFC2E37-119B-439D-82B6-49E9680EEA90}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "YamlDotNet.Samples", "YamlDotNet.Samples\YamlDotNet.Samples.csproj", "{689D2A30-C52E-4CF8-8BA1-3178F4EB245A}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "YamlDotNet.Samples", "YamlDotNet.Samples\YamlDotNet.Samples.csproj", "{689D2A30-C52E-4CF8-8BA1-3178F4EB245A}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "YamlDotNet.Test", "YamlDotNet.Test\YamlDotNet.Test.csproj", "{A9F67018-0240-4D16-A4EA-BCB780D0AF05}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "YamlDotNet.Test", "YamlDotNet.Test\YamlDotNet.Test.csproj", "{A9F67018-0240-4D16-A4EA-BCB780D0AF05}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "PerformanceTests", "PerformanceTests", "{9A64B652-30A5-4632-AF4D-716EF29C5250}"
 EndProject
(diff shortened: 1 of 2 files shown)


aaubry:
The SLN file format is such a pain :disappointed:

am11:
It's the top feature request upstream: https://github.com/microsoft/msbuild/issues/1730 😄

aaubry:
Since I've merged this PR, every time I open the solution, VS updates the GUIDs back to `9A19103F-16F7-4668-BE54-9A1E7A4F7556`. This is really annoying as it forces me to reset the changes to the file on every commit. I'm tempted to revert this change.

am11:
VS has problems with this SLN because it is mixed; has some project files written in SDK style, others are in old format. This causes this GUID update both ways around. Before this PR, we were getting the GUID update when we were adding/removing/unloading/reloading the projects. Also VS was unable to discover the tests, which was my main pain point.

I think it would be better if we do not keep our solution as mixed, but rather convert all the C# projects referenced by SLN to SDK style (or old style). After that VS will stop updating GUIDs.

aaubry:
As far as I can tell, all projects are SDK style. We converted them some time ago and I just checked again. Did I miss one ?

am11:
You are right, all projects were updated. The reason why VS is modifying the SLN is because we are missing one configuration in SLN for `{C0363835-93DB-4AF1-9E3C-0689C65D0FD7}`. So VS tries to add this delta:

```diff
                {BDD556C6-A2A3-46C2-B70A-CA86C16BC060}.Debug-AOT|Any CPU.ActiveCfg = Debug|Any CPU
                {BDD556C6-A2A3-46C2-B70A-CA86C16BC060}.Release|Any CPU.ActiveCfg = Release|Any CPU
                {BDD556C6-A2A3-46C2-B70A-CA86C16BC060}.Release|Any CPU.Build.0 = Release|Any CPU
+               {C0363835-93DB-4AF1-9E3C-0689C65D0FD7}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+               {C0363835-93DB-4AF1-9E3C-0689C65D0FD7}.Debug|Any CPU.Build.0 = Debug|Any CPU
+               {C0363835-93DB-4AF1-9E3C-0689C65D0FD7}.Debug-AOT|Any CPU.ActiveCfg = Debug|Any CPU
+               {C0363835-93DB-4AF1-9E3C-0689C65D0FD7}.Debug-AOT|Any CPU.Build.0 = Debug|Any CPU
                {C0363835-93DB-4AF1-9E3C-0689C65D0FD7}.Release|Any CPU.ActiveCfg = Release|Any CPU
                {C0363835-93DB-4AF1-9E3C-0689C65D0FD7}.Release|Any CPU.Build.0 = Release|Any CPU
                {91A1F4BC-65C0-42E6-B5FD-320A2D59AF71}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```

When VS touches the SLN, it updates the GUIDs. If we persist this delta, then close/open VS, it does not modify the file (as long as we do not add/remove projects which is not a common usecase). 

as far as i have understood, the reason for not having those configuration is by design? those test projects fail on TravisCI when using Mono runtime.

Should we find some other way to skip those tests on TravisCI mono leg without removing these configurations from SLN file, and checkin this four liner diff?

aaubry:
I've added the missing solution configuration and VS seems to be happy now. I don't know why the configurations were missing, but that wasn't intentional. Thanks for the help!

am11:
Thanks! Could you also re-enable Travis and AppVeyor CI in PR runs? I will thinking about starting working on v1.2 emitters support soon™️. :)
I tried it before in my fork but one mono leg was failing in TravisCI. If that failure persist, we can add the `\**\*.cs` sources glob pattern under a non-mono target `Condition` in that project.

aaubry:
I didn't disable Travis and AppVeyor. There was a change to appveyor.yml that caused PRs not to run but I have now reverted it. Regarding Travis, I don't know why it has ceased working.

am11:
I also had issue with TravisCI in one of my repository, that PRs are not triggering the build. Reading some online threads, it seems like something has changed and we need to revoke our tokens.

Following steps fixed my issue:
* Remove all *Travis* like entries from https://github.com/settings/applications
* Signout from travis-ci.com/.org
* Log on to travis-ci.com/.org with GitHub credentials and authorize

PR #477, opened 2020-03-30
LowerCaseNamingConvention added

Roadrunner67:
For those who prefer lowercase in their YAML regardles of casing on the class properties


aaubry:
Thanks. I've made the constructor private because it is not supposed to be used. It remains public on other naming conventions to avoid making breaking changes.

aaubry:
Release [8.1.1](https://www.nuget.org/packages/YamlDotNet/8.1.1) has just been pushed to NuGet. In a few minutes it should be indexed and available.

halamah:
@aaubry 
What's the point of having a lowercase naming convention instead of lets say CaseInsensitive naming convention?

Is there a case insensitive alternative at the moment?

PR #479, opened 2020-04-03
Bugfix: Tag resolving for "int" and "float"

bernd5:
Missing support for 10.3.2. Tag Resolution

-> only null and bool were "resolved" correctly

### DIFF
--- a/YamlDotNet/Serialization/NodeTypeResolvers/TagNodeTypeResolver.cs
+++ b/YamlDotNet/Serialization/NodeTypeResolvers/TagNodeTypeResolver.cs
@@ -46,12 +46,12 @@ namespace YamlDotNet.Serialization.NodeTypeResolvers
             }
 
             //Null and bool are handled special (can be ignored here)
-            AddRegExp("[-+]?[0-9]+", "int");
-            AddRegExp("0o[0-7]+", "int");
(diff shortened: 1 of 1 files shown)


pensono:
Is there a test case to exercise this functionality?

bernd5:
If you deserialize a yaml file with following content, the values were interpreted not as numbers (you got it as string):

```Yaml
SomeIntValue: 42
SomeFloatValue: 12.1
```

Please have a look at [YamlSpec 10.3.2. Tag Resolution](https://yaml.org/spec/1.2/spec.html#id2805071).

Boolean and Null value deserilization was done fine already.

aaubry:
The section of the spec that you mention refers to the "Core" schema, which is not the default one that parser should provide. I don't think this should be the default behaviour. Once we implement proper support for schemas, we'll be able to select the "Core" schema and have the behaviour you desire.

bernd5:
What is the default one that a parser should provide?

If you deserialize this yaml file and serialize it back you loose the type info. You get: 

```YAML
SomeIntValue: "42"
SomeFloatValue: "12.1"
```

In addition you could argue for bool and null equally. You could deserialze those values to the strings "" or "true", "false"...
But this would be very bad...

EdwardCooke:

Issue #271, opened 2017-07-28
YamlIgnoreAttribute can not apply to overrides


qaqz111:
```C#
public abstract class CfgFileBase {
	[YamlIgnore]
	public virtual string Path {get;set;} = @"C:\settings.yaml";
	[YamlIgnore]
        public virtual xxx xxx1 {get;set;}

	//...

	[YamlIgnore]
        public virtual xxx xxx20 {get;set;}
}

public class CfgFileForClass1 : CfgFileBase {
	public override string Path {get;set;} = @"D:\c1.test\settings.yaml";
	//...
}

//...

public class CfgFileForClass100 : CfgFileBase {
	public override string Path {get;set;} = @"D:\c100.test\settings.yaml";
	//...
}
```

While serializing `CfgFile` to yaml, the overriden `Path` was write to the output which is not desired.

Yes I can give a `[YamlIgnore]` to overridden property, but what if I have 100 or more derived class and 10 or more properties per class have to attach a `[YamlIgnore]`?

It's better to append a property `ApplyToOverrides` to indicate this attribute should apply to overrides, like `ScriptIgnore` in `System.Web.Script.Serialization`:
[`ApplyToOverrides` of `ScriptIgnore` in `System.Web.Script.Serialization`](https://docs.microsoft.com/en-us/dotnet/api/system.web.script.serialization.scriptignoreattribute.applytooverrides)

aaubry:
I agree that this needs fixing. Is is necessary to add a property to control this behavior? Is there any use case where one would not want the `YamlIgnore` to apply to an overriden property?

qaqz111:
I think it is very rarely that one would not want the `XXXIgnore` to apply to an overriden property while using a serializer library, so as you can see the `ApplyToOverrides` of `ScriptIgnore` is default to `true`, but provided a way for user to choose not to apply it to an overriden.

This pattern can be used by `YamlDotNet` as well. So, my suggestion is to add a property to control this behavior and set it's default value to allow attribute to apply to overrides.

EdwardCooke:
I just verified this issue is resolved with the following code and result:
```csharp
using YamlDotNet.Serialization;

var serializer = new SerializerBuilder()
    .Build();

var test = new Y
{
    NotIgnored = "hi",
    Ignored = "bye"
};
Console.WriteLine(serializer.Serialize(test));


public class X
{
    public string NotIgnored { get; set; }

    [YamlIgnore]
[...]

PR #481, opened 2020-04-18
YamlIgnore attributes are inherited

pensono:
The source of this bug is the fact that IMemberInfo.GetCustomAttributes actually ignores its "inherit" parameter.

https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.getcustomattributes

This addresses #271

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -413,6 +413,25 @@ namespace YamlDotNet.Test.Serialization
         }
     }
 
+    public class IgnoreExampleBase
+    {
+        [YamlIgnore]
+        public virtual String IgnoreMe
+        {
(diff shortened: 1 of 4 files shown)


aaubry:
This does not build on netstandard1.3 because `Attribute.GetCustomAttributes` is not available. Do you know of an alternative for that platform ?

pensono:
I've implemented a workaround for netstandard1.3.

If this workaround can't be used there's also the possibility of adding netstandard2.0 as a target for netstandard1.3 builds. There is some documentation that suggests this: https://github.com/dotnet/standard/blob/master/docs/versions.md#how-do-i-know-which-net-standard-version-i-should-target

aaubry:
This looks good. I'd rather avoid changing the targets at this point because that could break other people's code. I have no way of knowing which target platforms are actually used.
Thanks for the fix!

PR #482, opened 2020-04-24
Add support for private properties.

willson556:
Motivation: It can be necessary to serialize and deserialize `internal` (or more private) properties in order to restore object state. It is desirable to restrict access to these properties to encapsulate functionality.

Extend the `ReadablePropertiesTypeInspector` to handle non-public properties.

While it is possible for the user to provide custom type inspectors, it's not easy for these to surface additional properties. An alternative approach to the work here would be to make `ReflectionPropertyDescriptor` protected and unseal `ReadablePropertiesTypeInspector`. The  BuilderSkeleton` would also need to allow the root `ITypeInspector`(s) to be customized.

This approach surfaces the non-public properties while leaving any more complex filtering (such as only serializing them if they have an extra attribute) to downstream type inspectors.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -539,4 +539,32 @@ namespace YamlDotNet.Test.Serialization
         [YamlIgnore]
         public string fourthTest { get; set; }
     }
+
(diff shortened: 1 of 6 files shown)


aaubry:
I'm sorry it took so long to review this PR, but thanks anyway!

PR #483, opened 2020-04-27
Allow values after DocumentStart token

am11:
Mappings are not allowed after document start token, however, values are allowed. The error case checks were little too strict in that area. Removed the checks and added positive and negative tests to that effect.

Also fixed the spec URL in the README. Asymmetry is known: https://github.com/yaml/yaml-test-suite/issues/66.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ YamlDotNet's conformance with YAML specifications:
 
 |            YAML Spec                | YDN Parser | YDN Emitter |
 |:-----------------------------------:|:----------:|:-----------:|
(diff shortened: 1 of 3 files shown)


am11:
@aaubry, TravisCI is now running with PRs but AppVeyor is not. Maybe we can revoke and reauthorize AppVeyor.

This case was only affecting when document start is followed by value on the first line. Subsequent lines tests are covered by spect [27NA](https://github.com/yaml/yaml-test-suite/tree/data-2020-02-11/6LVF) and [6LVF](https://github.com/yaml/yaml-test-suite/tree/data-2020-02-11/6LVF) etc.

PR #486, opened 2020-04-28
Fix use of `DynamicTypeResolver` during deserialization.

willson556:
Previously the type resolver would only be able to return the static
type because the object constructed by the factory was not passed into
the type descriptor.

See the added tests for a couple of use cases.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -40,6 +40,7 @@ using YamlDotNet.Serialization;
 using YamlDotNet.Serialization.NamingConventions;
 using YamlDotNet.Serialization.ObjectFactories;
 using YamlDotNet.Serialization.TypeInspectors;
+using YamlDotNet.Serialization.TypeResolvers;
 
 namespace YamlDotNet.Test.Serialization
 {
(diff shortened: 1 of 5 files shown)


YamlDotNet.Test/Serialization/SerializationTests.cs
```diff
@@ -414,6 +415,7 @@ public void RoundtripDerivedClassWithSerializeAs()
                 obj,
                 new SerializerBuilder()
                     .WithTagMapping("!InheritanceExample", typeof(InheritanceExample))
+                    .WithTypeResolver(new StaticTypeResolver())
```
willson556, line 418: I'm a little worried about this change --  it was required to make the test pass. Without it, extra fields from the `Derived` type would be emitted.

I'm definitely open to suggestions for other ways to make the below tests pass.

aaubry:
I'm not sure that this change is correct. What are you trying to achieve ?
I don't think that the type inspectors should use the type resolver at that point.

willson556:
> 
> 
> I'm not sure that this change is correct. What are you trying to achieve ?
> I don't think that the type inspectors should use the type resolver at that point.

We're trying to achieve a couple of things (both demonstrated in the added tests):
- Deserialize properties that aren't present in the expected/static type but are in the instantiated type.
- Deserialize properties which are get-only in the expected/static type but have setters in the instantiated type.

EdwardCooke:

willson556:
That's fine; I'm no longer working on the project utilizing YamlDotNet. The project moved forward with a fork that had these changes.

Issue #484, opened 2020-04-27
Tracking issue for YAML 1.2 emitter support


am11:
YDN v8.1.1+ support  scanning and parsing of YAML 1.2 documents by the spec. This is indicated in the repo's README.

This issue is to track support for YAML 1.2 for emitter and (JSON) serializer.

As a prerequisite cleanup, I propose to refactor some bits for better diagnosability and facilitating emitter spec testing:

- [ ] Move libyaml event stream in its own emitter, so existing parser spec test runner and [yaml-runtime](https://github.com/yaml/yaml-runtimes/blob/841391aec8d982a4a598927fd3786fdabc36e19f/docker/dotnet/utils/event/Program.cs#L26) can share it: https://github.com/aaubry/YamlDotNet/blob/7f8c2992ee2b3e627e5719ba4d892a7ce209c68a/YamlDotNet.Test/Spec/SpecTests.cs#L116
- [ ] Extract spec suite directory discovery into utilities.
- [ ] Rename `SpecTests.cs` to something like `ParserSpecTests.cs`.

after that:
- [ ] Add `EmitterSpecTests.cs` with execution, assertions and ignore lists.

After the initial bootstrapping of emitter test runner, add the list of failing tests below this line for tracking.

`TODO: add list`

aaubry:
This looks good. FYI, I'm currently working on implementing [schemas](https://yaml.org/spec/1.2/spec.html#Schema), based on the 1.2 spec. This may be relevant for this work.

am11:
Cool, looking forward to it. 🙂
I haven't' looked much deeper but I think we have some spec tests validating schema on emitter side.

After the #485 refactoring, I will switch to working on emitter spec tests execution.

bootzin:
How's the progress on this?

am11:
@bootzin, spec suite for emitters was configured in master branch, such that anyone can contribute to improve the coverage and de-ignore the related specs. https://github.com/aaubry/YamlDotNet/blob/e32f92ed079c1b62fdd949b37c23db047649934c/YamlDotNet.Test/Spec/SerializerSpecTests.cs#L39

[...]

PR #491, opened 2020-05-10
Disallow usage of alias before anchor declaration

am11:
According to [YAML 1.1 `§8.4`](https://yaml.org/spec/1.1/#id902561) and [YAML 1.2 `§7.1`](https://yaml.org/spec/1.2/spec.html#id2786196):

> It is an error to have an alias node use an anchor that does not previously occur in the document.

Currently we emit `a: test0`  for this YAML:

```yaml
a: *anchor1
b: &anchor1 test0
```

PR aligns it with spec and changes it to throw `AnchorNotFoundException` instead.

Adjusted some tests which were asserting that forward alias usage is allowed.

Also added a test for duplicate anchors case, explicitly asserting what values are expected, per `YAML 1.2 §3.2.2.2`: https://github.com/yaml/yaml-spec/issues/48.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -234,19 +234,6 @@ namespace YamlDotNet.Test.Serialization
         }
 
         [Fact]
-        public void DeserializationOfObjectsHandlesForwardReferences()
-        {
-            var text = Lines(
-                "Nothing: *forward",
-                "MyString: &forward ForwardReference");
-
-            var result = Deserializer.Deserialize<Example>(UsingReaderFor(text));
-
-            result.ShouldBeEquivalentTo(
-                new { Nothing = "ForwardReference", MyString = "ForwardReference" }, o => o.ExcludingMissingMembers());
-        }
-
-        [Fact]
         public void DeserializationFailsForUndefinedForwardReferences()
         {
             var text = Lines(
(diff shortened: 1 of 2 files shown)


aaubry:
I know that this is against the spec. When I implemented support for forward references, I hadn't understood correctly that part of the spec and assumed that the anchor could appear after it was referenced.
The problem is that discontinuing this support would be a breaking change. I don't know if anyone is relying on this, and I agree that we should do it, but that will need to be included in a major release.
Also, there's more code that should be removed in the deserialization pipeline, namely everything that is related to `IValuePromise`. That's the abstraction that allows to deserialize an object and later set the properties that were resolved later.

am11:
I agree that it would be a breaking change among other required breakages which we will need to fully implement #484 (both serialization and deserialization conforming to spec v1.1 and v1.2). Most of the emitter spec conformance will take place step by step due to the nature of changes required to fit into the current API design. IMO, other than the required cleanups along the way, the rest of the cleanups, refactorings and performance improvements can be done once the correctness (standard conformance) is completed.

Is it a good time to target the next release as 9.0 and wait for all the changes we need (as long as it takes to accomplish spec v1.2 feat)? Otherwise, if there is a plan to ship some important fixes in v8.x, I think we can branch out v9.0 and push these changes there.

aaubry:
Sorry, I forgot to follow up on this conversation. I think there's no problem in merging this now. There's no need to cleanup the `IValuePromise` stuff now as I have already discontinued that in the schemas branch.
Regarding the releases, I don't really have a plan. I just increase the version numbers according the changes that were performed.

PR #497, opened 2020-05-28
Fix indentation check for first content line

### LINKED ISSUE #492: 8.1.1 bug
test yaml: 

    Text: >
       Sed ut perspiciatis, 
       maiores alias.   
   
    Value: foo

test code:

    static void Main(string[] args)
    {
        var deserializer = new DeserializerBuilder().WithNamingConvention(PascalCaseNamingConvention.Instance).Build();
        using (var textReader = File.OpenText("z:/test.yaml"))
        {
            var obj =  deserializer.Deserialize<Data>(textReader);
        }
    }
    public class Data
    {
        public string Text { get; set; }
        public string Value { get; set; }
    }

YamlDotNet 8.1.0 - works fine
YamlDotNet 8.1.1 cause error:

    YamlDotNet.Core.SemanticErrorException: '(Line: 5, Col: 1, Idx: 62) - (Line: 5, Col: 1, Idx: 62): 
    While scanning a literal block scalar, found more spaces in lines above first content line.'

.net48/netcore3.1

am11:
When the first content line has more spaces than the preceding non-content lines, it is an error (spec test S98Z).

However, this only holds if that first content line belongs to the multi-line scalar. This PR fixes the scanner to identify that case.

Fixes #492.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -194,6 +194,18 @@ namespace YamlDotNet.Test.Serialization
         }
 
         [Fact]
+        public void DeserializeWithGapsBetweenKeys()
+        {
+            var yamlReader = new StringReader(@"Text: >
+  Some Text.
+  
+Value: foo");
+            var result = Deserializer.Deserialize(yamlReader);
+
+            result.Should().NotBeNull();
+        }
+
+        [Fact]
         public void SerializeCustomTags()
         {
             var expectedResult = Yaml.StreamFrom("tags.yaml").ReadToEnd().NormalizeNewLines();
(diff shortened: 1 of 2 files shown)


am11:
@aaubry, some of these cases slipped when we were attempted to tighten the net. There is a test gap in some areas. I was thinking about running some other test suites from Python, JavaScript implementations, at least locally. However, that is a bit non-trivial as they also have false-positives and it requires re-parsing the specs. I will try to do that at some point.

Meanwhile, I think we can take these bug reports as a way to improve our implementation.

aaubry:
Thanks al lot! Sorry that I had to mention you, but you really have much more context than me in this area. It is normal that errors are introduced sometimes. It shows that the official test suite is missing some tests.

Issue #511, opened 2020-07-15
Exception if string is indented by more spaces than the MaxBufferLength in Scanner.cs


stuartbean:
``` yaml
Literal: |2-
          <hr>
```
This yaml (valid as far as I can tell) causes exception in Scanner.cs due to the following code exceeding the buffer:

``` csharp
                // Have we find a non-empty line?

                if (!analyzer.IsBreak())
                {
                    if (isLiteral && isFirstLine == true)
                    {
                        int localIndent = cursor.LineOffset;
                        int i = 0;
                        for (; !analyzer.IsBreak(i) && analyzer.IsSpace(i); ++i, ++localIndent) ; // BUG HERE
                        if (analyzer.IsBreak(i) && localIndent > cursor.LineOffset)
                        {
                            isFirstLine = false;
                            indentOfFirstLine = localIndent;
                        }
                    }
                    break;
                }
```

I cannot work out how to modify this code to achieve its original intent and protect against exceeding the buffer overrun. For now I've simply added a limit check in the for loop. I'll put up a PR with the unit test and "fix".

PR #512, opened 2020-07-15
Prevent buffer overrun for heavily indented literal strings

stuartbean:
Unit test / fix for issue #511 

I'm unsure how to implement a fix that retains the original intent of that segment of code without being able to grab more content as the buffer runs out. In the meantime I've added a limit check to at least prevent the exception.


aaubry:
I'm sorry for not answering sooner. This change no longer works because the `MaxBufferLength` constant has been removed. Now the buffer has always 1024 bytes, so while the issue is still there, it's harder to trigger it.
Maybe we should put back the constant so that your check will work, what do you think ?

stuartbean:
I've executed my unit test and it passes with the latest code base so I'm inclined to leave it at that. Thanks for the response and for providing the package in the first place - excellent work.
________________________________

aaubry:
I'm closing this as it seems that the issue has been fixed separately.

PR #516, opened 2020-08-07
Add simple quickstarts onto the home page to show where to start

sonic1981:
I've added some simple quick starts to your Readme. I was missing this when picking this library up. I did find the examples but they were a bit buried.


aaubry:
Thanks!

PR #522, opened 2020-08-22
docs: Correction to Quick Starts in README.md

cduplantis:
In README.md, the serialization and deserialization quick starts perform the opposite operations than what is declared. (They are backwards).

I took the opportunity to flush out the examples a tad bit more and provide a working .NET Fiddle to them. https://dotnetfiddle.net/CQ7ZKi.


aaubry:
Thanks!

Issue #387, opened 2019-02-27
Serializer misconverts "null" string to null value, "true" string to true value


justinyoo:
Hi, Team.

I'm using YamlDotNet v5.4.0, which is the latest version at the time of this writing. I have a very simple JSON payload like:

```json
{
  "hello": "null",
  "world": "true"
}
```

As you can see, both `"null"` and `"true"` are string, not literally `null` or `true` value. Therefore, if I converted that JSON into YAML, the YAML file should look like:

```yaml
hello: "null"
world: "true"
```

However, unlike the expectation, it actually returns:

```yaml
  hello: null,
  world: true
```

How can I prevent those values from being converted, but left as string literal? I used the code like:

```csharp
public static string ToYaml(this string json)
{
    if (string.IsNullOrWhiteSpace(json))
    {
        return null;
    }

    var converter = new ExpandoObjectConverter();
    ExpandoObject deserialised;

    try
    {
        deserialised = JsonConvert.DeserializeObject<ExpandoObject>(json, converter);
    }
    catch (Exception ex)
    {
        throw new InvalidJsonException(ex.Message);
    }

    var yaml = new SerializerBuilder().Build().Serialize(deserialised);
}
```

Obviously the last line `var yaml = new SerializerBuilder().Build().Serialize(deserialised);` doesn't return the correct value.

Is there a workaround to prevent this?

justinyoo:
The original issue was raised here: https://github.com/TeamYARM/YARM-CLI/issues/1

justinyoo:
I think YAML allows string value with/without the double quotation marks. If YamlDotNet forces to use double quotes for the string values, it could resolve this issue.

aaubry:
It seems that this behavior could be improved, but I'll need some more time to look into it.

erndob:
> It seems that this behavior could be improved, but I'll need some more time to look into it.

Are you actively working on it? If you could provide some initial guidance where to start looking, I can try to work on this myself.

aaubry:
I'm not currently working on this, but I'm not sure either where to start implementing this, sorry.

kellypleahy:
I've been looking at this a bit to try to come up with a reasonable solution, as I've had this same issue in my usage of YamlDotNet and am using it for data that I don't have control over so I really either (a) need it to work, or (b) use a different serialization library / format.

I definitely prefer (a) of those two options..., so I thought I'd see if I could provide a patch.

Here's my line of thinking on this:
[...]

PR #532, opened 2020-09-19
Feature/round trip output (special) strings (i.e. 'null', etc.) as strings.

kellypleahy:
Hi, not sure if this meets your requirements for pull requests, so feel free to tell me what I need to improve to do so.

I was having trouble building all the tests / running all the tests so it's possible these don't work on all platforms - for some reason my machine couldn't run tests on many of the platforms (but I was able to run on .NET core platforms and the tests work there).

If you'd like to talk about what I need to do to get my setup working, I'd be happy to do so and do additional work to bring this pull request in line.

This pull request relates to issue #387 and I believe fixes the associated issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ YamlDotNet/Properties/AssemblyInfo.Generated.cs
 
 /.vs
 /YamlDotNet/Properties/AssemblyInfo.cs
+
+/.idea
--- a/YamlDotNet.Test/Core/EmitterTests.cs
+++ b/YamlDotNet.Test/Core/EmitterTests.cs
@@ -346,6 +346,39 @@ namespace YamlDotNet.Test.Core
         }
 
         [Theory]
+        [InlineData("NULL")]
+        [InlineData("Null")]
+        [InlineData("null")]
+        [InlineData("~")]
+        [InlineData("true")]
+        [InlineData("false")]
+        [InlineData("True")]
+        [InlineData("False")]
+        [InlineData("TRUE")]
+        [InlineData("FALSE")]
+        [InlineData("0o77")]
+        [InlineData("0x7A")]
+        [InlineData("+1e10")]
+        [InlineData("1E10")]
+        [InlineData("+.inf")]
+        [InlineData("-.inf")]
+        [InlineData(".inf")]
+        [InlineData(".nan")]
+        [InlineData(".NaN")]
+        [InlineData(".NAN")]
+        [Trait("motive", "issue #387")]
+        public void StringsThatMatchKeywordsAreQuoted(string input)
+        {
+            var events = SequenceWith(
+                Scalar(input).ImplicitPlain
+            );
+
+            var yaml = EmittedTextFrom(StreamedDocumentWith(events));
+            yaml.Should()
+                .Contain(string.Format("- '{0}'", input));
+        }
+
+        [Theory]
         [InlineData("b-carriage-return,b-line-feed\r\nlll", "b-carriage-return,b-line-feed\nlll")]
         [InlineData("b-carriage-return\rlll", "b-carriage-return\nlll")]
         [InlineData("b-line-feed\nlll", "b-line-feed\nlll")]
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -154,6 +154,13 @@ namespace YamlDotNet.Test.Serialization
         }
 
         [Fact]
+        public void DeserializeQuotedStringNullAsString()
+        {
+            var result = Deserializer.Deserialize<string>(UsingReaderFor("'null'"));
+            result.Should().Be("null");
+        }
+
+        [Fact]
         public void RoundtripEnums()
         {
             var flags = EnumExample.One | EnumExample.Two;
--- a/YamlDotNet/Core/Emitter.cs
+++ b/YamlDotNet/Core/Emitter.cs
@@ -297,6 +297,21 @@ namespace YamlDotNet.Core
                 return;
             }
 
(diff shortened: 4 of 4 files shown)


aaubry:
Hi, thanks for taking the time to submit this PR!

It's weird that you weren't able to run all the tests. I would expect them to run out-of-the-box without issues. If you can share some information about that, I would like to look into it.

Regarding the actual changes, this is something that will be made obsolete by some work that is ongoing to properly implement schemas. Having said that, it is still valuable to have a workaround in the meantime. I just need some time to validate it on my end and then I will merge it.

aaubry:
I've tested these changes but they cause some tests to fail. I think the problem is that you don't have enough context inside the emitter to correctly pick a style. One option would be to move your changes into another implementation of `IEmitter` that would decorate another emitter to add this behaviour. That way, it would be opt-in, so we could have this feature without breaking the compliance with the spec.

kellypleahy:
OK, let me see if I can find some time over the next day or so to look at the tests.  I don't see why this would break the specs, as I can't ever imagine it being correct to emit a string as a "primitive value" that isn't a string, but now *I* don't have enough context to emit the right thing ;)

I'll let you know how it goes getting the tests / build working.

aaubry:
The tests break because you are assuming that plain strings are always to be interpreted as numbers or booleans or whatever matches, but that's not what the spec says. The spec specifies that how you resolve tags for plain scalars depends on the schema that is being used. For example, the "core" schema, defines that `true` should be interpreted as a boolean. But if you were using the "failsafe" schema, that would just be a string.
YamlDotNet doesn't implement this correctly because that scalar interpretation step is hard-coded during the deserialization, which makes it difficult to do the opposite in the emitter, because there's not enough context.
I'm working on implementing proper support for schemas to solve this problem in a clean way, but it's work in progress at the moment.

kellypleahy:
Interesting.  I thought my change would only affect the "output" side of the equation.  I wouldn't have expected outputting to ever want 'true' to be treated as boolean true but I guess I don't really understand the whole picture.

If you have any guidance on how to make this a "opt-in" feature, I'd be happy to do that.  I couldn't figure out a good place to have flags that control either (a) which emitter is used, or (b) options on the emitter itself.  It seems like if there was a way to get configuration to the emitter, it would be pretty easy to turn all this off with the "if" statement that triggers the behavior, I just didn't know how/where to put that configuration.

I would still like to get the tests running so I can see what you're talking about, but that's probably not until the weekend for me as this is a "side" project and I'm pretty busy this week.

Any specific tips on what you expect to (a) install, and (b) download from somewhere to get a "clean" machine building this?  I also generally don't use VS but use Rider instead, though I'd be fine to use VS (community edition) if I need to for this small bit of work.  My guess is I need some of the build frameworks and/or .NET SDKs as I only work in .NET core and probably don't have all the .NET framework proper stuff installed.

aaubry:
If you have .NET core installed, you can just run the netcore tests. The net46 tests are the same, and in most cases there should be no difference. This command should do it I think:
```c#
dotnet test --framework netcoreapp3.0
```

I had a docker image that contained all the required dependencies, but sadly it is failing due to some SSL error. I think the nuget version that it contains is outdated and does not support newer protocols.

If you want to setup a full environment, you'll need at least to install the .NET framework 4.6.1 and probably .NET framework 3.5 as well. You'll also need the latest .NET core SDK.

kellypleahy:
OK, thanks.  I'll try both the 'core only' route and the .NET SDK.

Where do you get the files that the repo is looking for for the test input?  I downloaded some, but I'm not sure I did it correctly as there weren't any instructions in the repo about this.

aaubry:
> If you have any guidance on how to make this a "opt-in" feature, I'd be happy to do that.
It depends. If you want to solve this for the serializer only, you could add an implementation of `IEventEmitter` (extend `ChainedEventEmitter`, it's simpler), simmilar to the existing `JsonEventEmitter`. You can override the `Emit(ScalarEventInfo eventInfo, IEmitter emitter)` method and you should be able to select the style of the scalar, when it is a string. The `ScalarEventInfo` should contain information about the .NET type that originated the scalar, so you'l know whether to quote or not.
You can register that class by calling `WithEventEmitter` on the `SerializerBuilder`. You can use the example from [this documentation](https://github.com/aaubry/YamlDotNet/wiki/Serialization.Serializer#witheventemitter) as inspiration.

aaubry:
> 
> 
> OK, thanks. I'll try both the 'core only' route and the .NET SDK.
> 
> Where do you get the files that the repo is looking for for the test input? I downloaded some, but I'm not sure I did it correctly as there weren't any instructions in the repo about this.

They come from submodules. You need to call `git submodule update --init`.

kellypleahy:
Ah, so you mean I could do this in "user code"?

Thanks re: submodules - I didn't think to even look for those!  I really wish git did a better job of mentioning them when you first clone (or maybe I am just dumb and missed it).

kellypleahy:
Antoine, this is what I came up with:

https://gist.github.com/kellypleahy/59635cbfce5d635ef6a8b66769c95748

but I don't seem to have a way to tell the difference between when something is the LHS of a object mapping (i.e. property name) and when something is the RHS (i.e. value).  Is there something I should be doing instead of the way I did this?

aaubry:
I'm sorry for not answering sooner. Unfortunately there's no easy way to know whether you are in the LHS vs RHS. That's a limitation of the API. You could detect that by having a state machine in the event emitter, but that's a lot of work...

EdwardCooke:
This is really old, with a lot of changes since this PR was made. I have used pieces of this for my next PR which will allow string quoting of special values. It will be an opt-in feature so it doesn't break anybody that is currently using this library, since starting to quote things could be an unexpected change to current users.
Closing this.

PR #533, opened 2020-09-21
Small performance improvements

aaubry:
Improves the performance of the following classes:
 - **InsertionQueue**: Reduces the amount of array copies to zero, in most typical cases;
 - **LookAheadBuffer**: Always read 1024 bytes at once instead of at most 4.


YamlDotNet/Core/Scanner.cs
```diff
@@ -109,7 +109,7 @@ public class Scanner : IScanner
         /// <param name="skipComments">Indicates whether comments should be ignored</param>
         public Scanner(TextReader input, bool skipComments = true)
         {
-            analyzer = new CharacterAnalyzer<LookAheadBuffer>(new LookAheadBuffer(input, MaxBufferLength));
+            analyzer = new CharacterAnalyzer<LookAheadBuffer>(new LookAheadBuffer(input, 1024));
```
am11, line 112: nit: we can set 1024 to `MaxBufferLength` or delete `MaxBufferLength` as it is unused.

aaubry: Fixed. Thanks for point it out.
That constant actually represented the minimum size that the buffer needs to have, but it is inefficient to use such a small buffer.

PR #539, opened 2020-10-24
Fix deserialization for Nullable types.

brendandburns:
Octal (and Hex and other format) deserialization assumes that the type is concrete, but it should also work for Nullable types.

#ref https://github.com/kubernetes-client/csharp/issues/500

Thanks!

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -138,6 +138,14 @@ namespace YamlDotNet.Test.Serialization
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


aaubry:
Thanks!

brendandburns:
Thanks for merging! Any ideas about when this will make it into a release?

aaubry:
I can create a release next thursday, 2020-10-29.

brendandburns:
Fantastic, that's awesome! many thanks!

aaubry:
I'm sorry I was unable to release when I expected to. I've just published release [9.1.0](https://www.nuget.org/packages/YamlDotNet/9.1.0) that includes this fix. Thanks for your patience!

PR #540, opened 2020-10-26
Allow block style in multiline scalars with trailing spaces

Himura2la:
We use a DocFX-based documentation system, which stores markdown in YAML. DocFX-Flavoured Markdown has a syntax for adding a `<br>` tag, and [it looks as follows](https://gist.github.com/shaunlebron/746476e6e7a4d698b373):

```markdown
line before br  <!-- notice the two spaces -->
line after br
```

this converts to 

```html
<p>
  line before br
  <br>
  line after br
</p>
```

When we store it in YAML and re-serialize the value, it's being converted from 

```yaml
summary: >-
  line before br  

  line after br
```
to 
```yaml
summary: "line before br  \nline after br"
```

, which is harder to read and edit. 

I see the reason why trailing spaces are disallowed in blocks (they are hard to notice, and single-line notation makes them more explicit), although I suggest you guys to review our case and compare pros and cons.
Thanks!

P.S.: I'll be grateful for the [hacktoberfest-accepted](https://hacktoberfest.digitalocean.com/hacktoberfest-update) label on this PR.

### DIFF
--- a/YamlDotNet.Test/Core/EmitterTests.cs
+++ b/YamlDotNet.Test/Core/EmitterTests.cs
@@ -208,6 +208,19 @@ namespace YamlDotNet.Test.Core
             yaml.Should().Contain(">");
         }
 
+
+        [Fact]
+        [Trait("motive", "pr #540")]
+        public void AllowBlockStyleInMultilineScalarsWithTrailingSpaces()
+        {
+            var events = SequenceWith(Scalar("hello  \nworld").ImplicitPlain);
+
+            var yaml = EmittedTextFrom(StreamedDocumentWith(events));
+
+            yaml.Should().Contain("\n");
+        }
+
+
         [Fact]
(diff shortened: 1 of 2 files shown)


Himura2la:
OK, I see it breaks three tests from yaml-test-suite:

- https://github.com/yaml/yaml-test-suite/blob/master/test/H2RW.tml
- https://github.com/yaml/yaml-test-suite/blob/master/test/T26H.tml
- https://github.com/yaml/yaml-test-suite/blob/master/test/DWX9.tml

It's strange, because the https://github.com/eemeli/yaml library is also being tested for compatibility with this suite, and passes the tests yet allowing trailing whitespaces in blocks. Maybe the current implementation needs more work...

Himura2la:
I hope it's ready, @aaubry, what do u think about it?

aaubry:
Sorry for the delay. I think this is fine, since it doesn't break any test I don't see any problem with this. Thanks!

PR #542, opened 2020-11-07
Fix the order of YamlMappingNode items

### LINKED ISSUE #414: YamlMappingNode does not preserve the order of the keys
When loading a YAML stream, the order of the key-value pairs should be preserved by YamlMappingNode.

tgillbe:
Fixes #414 and exposes an interface for inserting / deleting / accessing items by index.


aaubry:
Thanks!

PR #563, opened 2021-01-15
Fix imutable pattern in EmitterSettings.SkipAnchorName

FSDKO:
In class `EmitterSettings` the imutable pattern is not implemented consistent. `SkipAnchorName` is not handled correct.

The following test fails (this ist not a big thing):
```csharp
var settings1 = new EmitterSettings();
var settings2 = settings1.WithoutAnchorName();
Assert.IsFalse(settings1.SkipAnchorName);
```
Reason: The method `WithoutAnchorName()` does not create a new instance like all other Methods.

The following test also fails - and this is a trap:
```csharp
var settings = new EmitterSettings().WithoutAnchorName().WithBestIndent();
Assert.IsTrue(settings.SkipAnchorName);
```
Reason: all other Method don't give `SkipAnchorName ` into the ctor.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,14 +1,14 @@
 # Release notes
 
-## Release 9.1.2
+## Release 9.1.3
 
-- Fix writePtr decrementation in InsertionQueue [#558]
-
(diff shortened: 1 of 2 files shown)


aaubry:
Thanks!

PR #585, opened 2021-02-22
Enable filling get only ICollection property

ADIX7:
A get-only property can not be set but if it is an ICollection and the constructor initialized it we can copy the deserialized items with Add.

This yaml currently can't be deserialized to the class below.

```yaml
contents:
- asd1
- asd2
- asd3
- asd4
```

```csharp
public class Container
{
    public List<string> Contents { get; } = new List<string>();
}
```


aaubry:
I've reviews these changes and they seem fine. Could you add some tests to show that the feature works as intended? Thanks

aaubry:
Any news, @ADIX7 ?

ADIX7:
Sorry, I didn't had any time finish this.

EdwardCooke:

PR #595, opened 2021-03-10
Corrected link to Unity Asset store

grapefrukt:
This fixes the link to the Unity Asset store, as it seems to have broken. 
I'm not adding any unit tests for this change for obvious reasons.


aaubry:
Thanks!

Issue #594, opened 2021-03-09
YamlDotNet merges anchor list in wrong order.


Indomitable:
Hello,

According the documentation: https://yaml.org/type/merge.html 
> Keys in mapping nodes earlier in the sequence override keys specified in later mapping nodes

Considering this example:
```yaml
.first: &first
  x: 1
.second: &second
  x: 2

item:
  <<: [*second, *first]
```
element item.x should have value 2, but YamlDotNet produces value 1
Sample project can be found here: https://github.com/Indomitable/yaml-merge-bug
![image](https://user-images.githubusercontent.com/1831191/110525645-203b8500-8115-11eb-82c0-f41b3c1c19da.png)

BR

PR #596, opened 2021-03-10
Merge sequence of anchors in reverse order

Indomitable:
The idea is that we are adding anchor value after the sequence start instead after the anchors.

Fixes: #594 

- Remove the comment about the fact that it is not safe to fix it when anchors are defined later and added same test but with definition at the end of the document.

I'm only worried if there are consumers which actually reversed their sequence in order to resolve this bug, so we need to mark this in the release notes.


aaubry:
Thanks for this. I'll merge it now and release it soon.

Issue #597, opened 2021-03-19
How to handle cyclic references


210342:
I have asked similar question on StackOverflow, but I wonder if there is a bug with deserialization of models with cyclic references. [Link to the SO question](https://stackoverflow.com/questions/66713914/how-to-handle-cycles-in-object-graph-with-yamldotnet-serializers). Deserializing a previously serialized model with a circular reference throws an exception:
```
YamlDotNet.Core.AnchorNotFoundException: (Line: 7, Col: 5, Idx: 70) - (Line: 7, Col: 8, Idx: 73): Alias $o0 cannot precede anchor declaration
```
[.NET fiddle to reproduce](https://dotnetfiddle.net/7w6SM3)

210342:
I assumed this is a bug and proposed a fix in #598

PR #598, opened 2021-03-22
Create a value promise right after seeing an anchor

210342:
This PR is related to #597.

Creating a promise for an alias before deserialization allows the deserializer to handle cycles in object graph.

Currently, if there is a reference to an anchor of a node which is being deserialized, deserializer throws an exception `AnchorNotFoundException`.


aaubry:
Looks good. Thanks!

PR #608, opened 2021-05-04
Allow omitting empty collections via `DefaultValuesHandling`

premun:
**Motivation**
Sometimes, you cannot cannot afford to null your enumerables but you don't want them in your YAML.
This default value handling policy allows to prevent serialization of empty arrays/lists/dictionaries into
```yaml
value: []
```

**Change**
Added 2 new options: 
- `DefaultValuesHandling.OmitDefaultsOrEmpty`
- `DefaultValuesHandling.OmitNullOrEmpty`
which turns on omitting of empty collections.

### DIFF
--- a/YamlDotNet/Serialization/DefaultValuesHandling.cs
+++ b/YamlDotNet/Serialization/DefaultValuesHandling.cs
@@ -19,36 +19,34 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+using System;
+
 namespace YamlDotNet.Serialization
 {
     /// <summary>
     /// Specifies the strategy to handle default and null values during serialization of properties.
     /// </summary>
+    [Flags]
     public enum DefaultValuesHandling
     {
         /// <summary>
         /// Specifies that all properties are to be emitted regardless of their value. This is the default behavior.
         /// </summary>
-        Preserve,
+        Preserve = 0,
 
         /// <summary>
         /// Specifies that properties that contain null references or a null Nullable&lt;T&gt; are to be omitted. 
         /// </summary>
-        OmitNull,
+        OmitNull = 1,
 
         /// <summary>
-        /// More relaxed than OmitNull - omits null values and also collections/arrays/enumerations that are empty.
+        /// Omits collections/arrays/enumerations that are empty.
         /// </summary>
-        OmitNullOrEmpty,
+        OmitEmpty = 2,
 
         /// <summary>
         /// Specifies that properties that that contain their default value, either default(T) or the value specified in DefaultValueAttribute are to be omitted. 
         /// </summary>
-        OmitDefaults,
-
-        /// <summary>
-        /// More relaxed than OmitDefaults - omits default values and also collections/arrays/enumerations that are empty.
-        /// </summary>
-        OmitDefaultsOrEmpty,
+        OmitDefaults = 4,
     }
 }
--- a/YamlDotNet/Serialization/DefaultValuesHandling.cs
+++ b/YamlDotNet/Serialization/DefaultValuesHandling.cs
@@ -35,18 +35,18 @@ namespace YamlDotNet.Serialization
         Preserve = 0,
 
         /// <summary>
-        /// Specifies that properties that contain null references or a null Nullable&lt;T&gt; are to be omitted. 
+        /// Specifies that properties that contain null references or a null Nullable&lt;T&gt; are to be omitted.
         /// </summary>
         OmitNull = 1,
 
         /// <summary>
-        /// Omits collections/arrays/enumerations that are empty.
+        /// Specifies that properties that that contain their default value, either default(T) or the value specified in DefaultValueAttribute are to be omitted.
         /// </summary>
-        OmitEmpty = 2,
+        OmitDefaults = 2,
 
         /// <summary>
-        /// Specifies that properties that that contain their default value, either default(T) or the value specified in DefaultValueAttribute are to be omitted. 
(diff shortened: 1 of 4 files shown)


premun:
@aaubry would it be possible to get this looked at? It would help my other project a lot to have this capability

aaubry:
> @aaubry would it be possible to get this looked at? It would help my other project a lot to have this capability

I'm sorry, I wasn't able to look at this until today. I'm reviewing the PR now.

YamlDotNet/Serialization/ObjectGraphVisitors/DefaultValuesObjectGraphVisitor.cs
```diff
@@ -58,13 +59,29 @@ public override bool EnterMapping(IPropertyDescriptor key, IObjectDescriptor val
                     }
                     break;
 
+                case DefaultValuesHandling.OmitNullOrEmpty:
+                    if (value.Value is IEnumerable enumerable && !enumerable.GetEnumerator().MoveNext())
```
aaubry, line 63: The enumerable needs to be disposed after use.

YamlDotNet/Serialization/ObjectGraphVisitors/DefaultValuesObjectGraphVisitor.cs
```diff
@@ -58,13 +59,29 @@ public override bool EnterMapping(IPropertyDescriptor key, IObjectDescriptor val
...
                 case DefaultValuesHandling.OmitDefaults:
                     var defaultValue = key.GetCustomAttribute<DefaultValueAttribute>()?.Value ?? GetDefault(key.Type);
                     if (Equals(value.Value, defaultValue))
                     {
                         return false;
                     }
                     break;
+
+                case DefaultValuesHandling.OmitDefaultsOrEmpty:
+                    if (value.Value is IEnumerable ienumerable && !ienumerable.GetEnumerator().MoveNext())
```
aaubry, line 79: idem

YamlDotNet/Serialization/DefaultValuesHandling.cs
```diff
@@ -36,9 +36,19 @@ public enum DefaultValuesHandling
         /// </summary>
         OmitNull,
 
+        /// <summary>
+        /// More relaxed than OmitNull - omits null values and also collections/arrays/enumerations that are empty.
+        /// </summary>
+        OmitNullOrEmpty,
```
aaubry, line 42: Maybe it would be better to convert this enum to a `[Flags]`, so that it is easier to combine the options. If we need to add a new option later, we will need to add all combinations. Using flags avoids it. I would suggest the following structure:
```c#
[Flags]
public enum DefaultValuesHandling
{
    /// <summary>
    /// Specifies that all properties are to be emitted regardless of their value. This is the default behavior.
    /// </summary>
    Preserve = 0,

    /// <summary>
    /// Specifies that properties that contain null references or a null Nullable&lt;T&gt; are to be omitted. 
    /// </summary>
    OmitNull = 1,

    /// <summary>
    /// Specifies that properties that that contain their default value, either default(T) or the value specified in DefaultValueAttribute are to be omitted. 
    /// </summary>
    OmitDefaults = 2,

    /// <summary>
    /// Specifies that properties that that contain collections/arrays/enumerations that are empty. are to be omitted.
    /// </summary>
    OmitEmptyCollections = 4,
}
```

YamlDotNet/Serialization/ObjectGraphVisitors/DefaultValuesObjectGraphVisitor.cs
```diff
@@ -58,13 +59,29 @@ public override bool EnterMapping(IPropertyDescriptor key, IObjectDescriptor val
                     }
                     break;
 
+                case DefaultValuesHandling.OmitNullOrEmpty:
+                    if (value.Value is IEnumerable enumerable && !enumerable.GetEnumerator().MoveNext())
+                    {
+                        return false;
+                    }
+
+                    goto case DefaultValuesHandling.OmitNull;
```
aaubry, line 68: It would be cleaner to test this separately. If the enum is changed to flags, I this this will happen naturally.

aaubry (commented):
I think this feature is interesting, but requires some small design changes so that we don't get stuck in the future.

YamlDotNet/Serialization/ObjectGraphVisitors/DefaultValuesObjectGraphVisitor.cs
premun: Good catch. I had to go a more complicated way because the non-generic `IEnumerable` doesn't return `IDisposable` enumerator.

YamlDotNet/Serialization/DefaultValuesHandling.cs
premun: Yeah, makes more sense!

aaubry:
Thanks and sorry about the slow response time.

aaubry:
This feature has been released in [version 11.2.0](https://github.com/aaubry/YamlDotNet/releases/tag/v11.2.0).

premun:
Thanks @aaubry!!

tjstjsgt:
why is OmitEmptyCollections exclusive to OmitNull and/or OmitEmpty? It seems like these are two separate things. Right now I can't omit both empty collections AND null values, I have to choose one. Based on the comments it looks like that was the original plan but not implemented that way.

premun:
@tjstjsgt we have made these options into `[Flags]` as per this comment: https://github.com/aaubry/YamlDotNet/pull/608#discussion_r642061309

So you could set the policy to `DefaultValuesHandling.OmitNull | DefaultValuesHandling.OmitEmpty` which should do the trick.

Or is there a problem?

PR #613, opened 2021-05-26
Added the deserialize fix and test for the UTC DateTime

gjdonkers:
Deserialize the datetime to correct timezone as described on the Microsoft documentation for the "o" Roundtrip

```
// The example displays the following output:
//    6/15/2009 1:45:30 PM (Unspecified) --> 2009-06-15T13:45:30.0000000
//    6/15/2009 1:45:30 PM (Utc) --> 2009-06-15T13:45:30.0000000Z
//    6/15/2009 1:45:30 PM (Local) --> 2009-06-15T13:45:30.0000000-07:00
```

https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#Roundtrip

### DIFF
--- a/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
+++ b/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
@@ -91,7 +91,10 @@ namespace YamlDotNet.Serialization.NodeDeserializers
 
                 case TypeCode.DateTime:
                     // TODO: This is probably incorrect. Use the correct regular expression.
-                    value = DateTime.Parse(scalar.Value, CultureInfo.InvariantCulture);
+                    var style = scalar.Value.EndsWith("Z")
+                        ? DateTimeStyles.AdjustToUniversal | DateTimeStyles.AssumeUniversal
+                        : scalar.Value.Length > 27 ? DateTimeStyles.AssumeLocal : DateTimeStyles.None;
+                    value = DateTime.Parse(scalar.Value, CultureInfo.InvariantCulture, style);
                     break;
 
                 default:
--- a/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
+++ b/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
@@ -91,10 +91,7 @@ namespace YamlDotNet.Serialization.NodeDeserializers
 
                 case TypeCode.DateTime:
                     // TODO: This is probably incorrect. Use the correct regular expression.
-                    var style = scalar.Value.EndsWith("Z")
-                        ? DateTimeStyles.RoundtripKind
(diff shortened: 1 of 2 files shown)


YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
```diff
@@ -91,7 +91,10 @@ bool INodeDeserializer.Deserialize(IParser parser, Type expectedType, Func<IPars
 
                 case TypeCode.DateTime:
                     // TODO: This is probably incorrect. Use the correct regular expression.
-                    value = DateTime.Parse(scalar.Value, CultureInfo.InvariantCulture);
+                    var style = scalar.Value.EndsWith("Z")
+                        ? DateTimeStyles.AdjustToUniversal | DateTimeStyles.AssumeUniversal
+                        : scalar.Value.Length > 27 ? DateTimeStyles.AssumeLocal : DateTimeStyles.None;
```
aaubry, line 96: Testing the length seems a bit fragile. Maybe a regex would be more robust?

gjdonkers: I don't have the skills to create a good regex for this, so I don't know if it would be more robust. The previous implementation is still the same, `DateTimeStyles.None` is the default. And the 'o' roundtrip is compliant with the ISO 8601, so I thought this won't break other implementations but would only add the correct TimeZone if this is known.

gjdonkers: I know you are a busy man at the moment but is there a chance we can resolve this? What kind of regex are you thinking of?

aaubry: I'm sorry, I haven't had much free time. I made some experiments and it seems that instead of forcing the adjustment, you can simply pass the `DateTimeStyles.RoundtripKind` value to the style and you will get the correct kind. Have you tried that instead?

gjdonkers: Ah.. I've overlooked that one, that is definitely a better solution.

aaubry: I think you can always use the `DateTimeStyles.RoundtripKind`, without any test:
```c#
value = DateTime.Parse(scalar.Value, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
```

aaubry:
Thanks for this and sorry about the slow response time.

gjdonkers:
No problem. I’m glad it’s fixed.

gjdonkers:
One question, when do you think this will be released to the latest nuget?

aaubry:
This feature has been released in [version 11.2.0](https://github.com/aaubry/YamlDotNet/releases/tag/v11.2.0).

PR #624, opened 2021-07-15
Include start / end in exceptions thrown by type inspector

jairbubbles:
- Make sure to include `Start` / `End` in exception thrown by type inspector
- Use message from the serialization exception directly

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -1204,6 +1204,13 @@ y:
             var text = Lines("aaa: hello", "bbb: world");
             var actual = Record.Exception(() => Deserializer.Deserialize<Simple>(UsingReaderFor(text)));
             Assert.IsType<YamlException>(actual);
+            ((YamlException)actual).Start.Column.Should().Be(1);
+            ((YamlException)actual).Start.Line.Should().Be(2);
+            ((YamlException)actual).Start.Index.Should().Be(12);
+            ((YamlException)actual).End.Column.Should().Be(4);
+            ((YamlException)actual).End.Line.Should().Be(2);
+            ((YamlException)actual).End.Index.Should().Be(15);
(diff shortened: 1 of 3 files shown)


aaubry (changes requested):
I think including the location of the error is a useful improvement, however I don't think that changing `ITypeInspector` to take a scalar is a good idea for the following reasons:
- That interface should not be coupled to YAML structure. It is an abstraction over the .NET type system. In principle it could be used in contexts that have nothing to do with YAML.
- This is a breaking change that will force anyone who implements that interface to change their code. `ITypeInspector` is a frequently used extension point, so the impact would be high.

Instead I would suggest to catch the exception(s) and wrap them. In principle, this would also be a breaking change, although I believe that the impact would be quite low as I don't expect code to rely on these exceptions specifically.

What do you think?

jairbubbles:
@aaubry As it was not massively used in the solution I thought that maybe it was an internal interface.I tried my luck!  ☺

By the way there is this nice analyzer to detect breaking changes in the public API: 
https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md

Please note that I'm currently using `YamlDotNet` to create a linter for a .yml file and appart from the issue fixed here, it would have been nice to receive multiple serialization exceptions. For example you have a missing property at the beginning of the file and a syntax error at the end. 

We could accumulate the serialization errors and wrap them in a new exception type like `MultipleYamlException`.  What do you think @aaubry?

jairbubbles:
Hi @aaubry, I changed the code accordingly to your remarks, what do you think about it?

aaubry:
Thanks and sorry for the delay!

jairbubbles:
Thx! Let me know when you release a version with the fix.

jairbubbles:
Hi @aaubry  any plans to release a version including this fix?

aaubry:

PR #630, opened 2021-08-16
Add .NET 5 support

am11:
This patch allows us to build project on machine where only .NET 5 is installed.

### DIFF
--- a/YamlDotNet/Helpers/OrderedDictionary.cs
+++ b/YamlDotNet/Helpers/OrderedDictionary.cs
@@ -136,14 +136,14 @@ namespace YamlDotNet.Helpers
             list.RemoveAt(index);
         }
 
-#if !(NETCOREAPP3_1)
+#if !NETCOREAPP3_1_OR_GREATER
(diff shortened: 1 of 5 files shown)


YamlDotNet/Helpers/OrderedDictionary.cs
```diff
@@ -136,14 +136,14 @@ public void RemoveAt(int index)
             list.RemoveAt(index);
         }
 
-#if !(NETCOREAPP3_1)
+#if !NETCOREAPP3_1 && !NET5_0
```
jairbubbles, line 139: Why not testing `NETCOREAPP` directly? There is also `NETCOREAPP3_1_OR_GREATER` if you need to be more specific.

am11: Good call. We are targeting netcoreapp3.1 and above (aside from netstandard and netframework). I have updated the conditions to use `NETCOREAPP3_1_OR_GREATER`.

EdwardCooke:
.net 5 is end of life now. .netstandard2.0, 2.1 and net60 are valid targets in the coming release. Closing this.

PR #631, opened 2021-08-16
Rev yaml-test-suite submodule and fix Unix build

am11:
* Update `yaml-test-suite` submodule to latest release
  * Some new parser tests came to light which need addressing. I will follow up on this
  * Emitter cases are continuing to pile up, I just added new ones to the list
* Fix failing tests on Unix due to platform differences
* Add .gitattributes to avoid checking in code with CRLF ending (which leaves the repo with inconsistent line-endings)

### DIFF
--- a/YamlDotNet.Test/Spec/ParserSpecTests.cs
+++ b/YamlDotNet.Test/Spec/ParserSpecTests.cs
@@ -39,17 +39,17 @@ namespace YamlDotNet.Test.Spec
 
         private static readonly List<string> ignoredSuites = new List<string>
         {
-            // no spec test is ignored as of https://github.com/yaml/yaml-test-suite/releases/tag/data-2020-02-11
+            "5T43", "JR7V", "NKF9", "CFD4", "U99R"
         };
 
         private static readonly List<string> knownFalsePositives = new List<string>
         {
-            // no false-positives known as of https://github.com/yaml/yaml-test-suite/releases/tag/data-2020-02-11
+            // no false-positives known as of https://github.com/yaml/yaml-test-suite/releases/tag/data-2020-08-01
         };
 
         private static readonly List<string> knownParserDesyncInErrorCases = new List<string>
         {
-            "5LLU" // remove 5LLU once https://github.com/yaml/yaml-test-suite/pull/61 is released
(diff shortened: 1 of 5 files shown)


YamlDotNet.Test/Spec/ParserSpecTests.cs
```diff
@@ -39,71 +39,71 @@ private sealed class ParserSpecTestsData : SpecTestsData
 
         private static readonly List<string> ignoredSuites = new List<string>
         {
-            // no spec test is ignored as of https://github.com/yaml/yaml-test-suite/releases/tag/data-2020-02-11
+            "5T43", "JR7V", "NKF9", "CFD4", "U99R"
```
am11, line 42: @aaubry, I have started working on fixing these new cases; got the first one fixed here https://github.com/am11/YamlDotNet/commit/594324737e7e3b1cedf69f6d68b6818b705d4ac0.

----

### Plain style scanning

From parsing architecture perspective, one significant issue is that we have `FetchPlainScalar` in Scanner class taking up a lot of responsibility and glues up a lot of mixed concerns (from all kinds of plain styles) in one place; which makes harder to comprehend, if not impossible.

In particular, [YAML 1.2 `§7.3.3`](https://yaml.org/spec/1.2/spec.html#id2788859) talks about plain style of flow scalars, that is more restrictive than the `$other` scalar types and other styles of flow scalars. For the other two flow styles (single and double quoted), we have a dedicated method `FetchFlowScalar(bool isSingleQuoted)`, but for plain style the control flow either resorts to `FetchValue`, or (like in case of 5T43's fix) to `FetchPlainScalar`. IMHO, we should try to refactor flow scalar's plain style by changing its method's signature to `FetchFlowScalar(bool isQuoted, bool isSingleQuoted)` and implement its rules there. It might result in a little code duplication but much cleaner and implementation would be easily relatable with spec the, 1:1.

Similarly, for other kinds of `plain` styles, they could be structured in their corresponding parent concern and relatable with spec.

----

However, I believe priority 1 is correctness. Refactorings / optimizations can wait. I just wanted to record this message as the pain of finding the problem and my wounds therefore, are fresh at the moment. 8-)
Maybe someone has time to undergo this (very targetted) refactoring. My hands are full with stuff like real life, work and tracking down why `JR7V` and other are failing..  🤣

PR #634, opened 2021-09-14
Asset store update

steinbitglis:
Hey Antoine!
This is not a pull request that I actually want you to merge anywhere,
but I'm in need of a little bit of code review. I've held back on updating
the Unity Asset Store version of YamlDotNet, because C# 8 features were being used
after version 6.1.2.   But now that Unity has a much upgraded compiler that
supports C# 8, I was making another attempt, just to stay on par with your
progress. Unity now supports .net standard 2.0. There were a few
things in YamlDotNet 11.2.1 that required .net standard 2.1, but I think it's
pretty close.

So what this is, is just me asking if you could review my branch unity_compatible,
and see if you think I broke anything. I don't have a proper IDE set up for
the project, and I'm not able to run the test suite. It would be a really big job
for me just to run the tests, but if it's easier for you, could you do a quick review
and maybe a sanity check (if I've misunderstood the .net standard 2.1 features
or something.). The sample code seems to run alright in Unity, but I haven't
checked the output properly.

If you think it looks ok, I'll go ahead and package it up for the asset store, and
probably also figure out which exact Unity version is recent enough for this updated
code.

PR #636, opened 2021-09-28
Attempt to add a Unity project to test compilation

aaubry:
Another take at #634


steinbitglis:
Thanks! That was quite convenient, I have to say.

I'm now working your changes into the package that I'm going to upload.
But I'm getting a lot of warnings (using Unity 2021.1.18f1).

The following classes seem to conflict with .net standard 2.0 libraries:

- ConcurrentDictionary
- IReadOnlyDictionary
- IReadOnlyList
- Lazy
- LazyBuilder

Then at the bottom of TypeConverter.cs, there seems to be a portion of code that is
not meant for Unity, but it's going to be if the user doesn't explicitly define the symbol "UNITY" in their project settings.

A lot of files seem to expect a "nullable context". This is easily fixable by adding a file 'csc.rsp' next to the assembly definition file,
with the line '-nullable:enable'.

At the end of all that, I was left with these two warnings:

> Assets\Plugins\YamlDotNet\RepresentationModel\YamlDocument.cs(61,18): warning CS8618: Non-nullable property 'RootNode' is uninitialized. Consider declaring the property as nullable.
> Assets\Plugins\YamlDotNet\Serialization\YamlAttributeOverrides.cs(97,35): warning CS8600: Converting null literal or possible null value to non-nullable type.

Should I turn off the warnings, or is there some change that should be done?

image

steinbitglis:
It seems that the following files and folders in Portability are sufficient, and compiles with only the two mentioned warnings.
image

steinbitglis:
I could also delete _DeconstructionExtensions.cs_ and _TargetFrameworkAttribute.cs_, and everything seems fine.

steinbitglis:
I have fixed every issue I could find and submitted a package to the asset store. I guess I spent most of the time updating the store images.
I'll try and put together a pull request against _unity-compatible_, so you can have a look, but I'll have to make dinner first now 😅

PR #639, opened 2021-10-03
Unity compatible

steinbitglis:
Hello!
These are roughly the changes that I did before I submitted my package to the Unity Asset Store.
I would love to get feedback on whether this is the right way to go or not. I suppose I made a mess of the Portability folder, but I think it's a little bit improved from a Unity 2021 perspective.

The unity QA team will have a few days with the package before releasing it on the asset store.


steinbitglis:
I probably broke something in _TypeConverter_, but a final solution can not rely on the 'UNITY' define flag. I don't know exactly what to do, since I couldn't find an appropriate define flag. For now I did manual removal for my package.

krisrok:
just yesterday i've added the [SharpYaml ](https://github.com/xoofx/SharpYaml) fork to a unity project and it just works, maybe that is an alternative?
(i've stumbled upon this issue because i'm currently trying to figure out if should have a look at YamlDotNet or keep SharpYaml)

steinbitglis:
Even though this was an open pull request, the work had been done and YamlDotNet on the Asset Store got updated as intended.

Issue #519, opened 2020-08-16
Parsing performance of large YAML files


PJB3005:
We're using YAML as a data format in our game, from data definitions to map files. We have a modestly sized map files now (that is at the time of writing, [~700 KiB uncompressed/38k lines](https://github.com/space-wizards/space-station-14/blob/8f1a74dcd1b8973a730ef6aeebe1f9f427886843/Resources/Maps/saltern.yml)).

I should note that we're *not* using deserialization features, JUST the AST (`YamlMappingNode` etc directly).

The parsing performance is really starting the hurt (currently taking about 10% of debug server startup time *just* spent inside YamlDotNet parsing the main map file) and, the kicker, 90% of *that* time is spent inside various collection ops like `Array.Copy`, `List.Insert` and `List.RemoveAt`.

<details>

<summary>Some profiling results</summary>

![image](https://user-images.githubusercontent.com/8107459/90338892-2fc67580-dfed-11ea-8cf3-35f25da967f1.png)

![image](https://user-images.githubusercontent.com/8107459/90338898-39e87400-dfed-11ea-82d1-5a57ce22f2b1.png)

</details>

Could this be optimized? I tried saving the map files in flow style instead (since from my looking at the code the reason there's so many list butchering is because of parsing issues with block style?) but it did not seem to help perf. In fact it seems slightly worse with flow style.

aaubry:
There's probably room for improvement. I would be interested in exploring this but I can't dedicate much time to performance right now because I already have another development in progress. I can probably provide some help if you guide this process, though.

I've looked at `InsertionQueue`, which appears in your screenshot, and I can see that there's a big TODO:
> // TODO: Use a more efficient data structure

Looking at the `Dequeue` method, it is clear that it is really inefficient since it always removes the first item from a `List`, meaning that every other element needs to be shifted. We could start by changing this implementation to use a more efficient algorithm.

aaubry:
Would you be able to provide an example of such a large YAML file ?

PJB3005:
> Would you be able to provide an example of such a large YAML file ?

[From our repo](https://github.com/space-wizards/space-station-14/blob/8f1a74dcd1b8973a730ef6aeebe1f9f427886843/Resources/Maps/saltern.yml)

aaubry:
Hello!
[...]

PR #647, opened 2021-11-13
Change Mark to be readonly struct

lahma:
**!! This is binary breaking change !!**

To reduce memory allocations and improve scanning speed (less GC) I've changed `Mark` to be a read-only struct. In parsing test code profiling showed that over 20% of memory allocations were from constructing `Mark` instances. This is a binary breaking change, but I'd argue the performance benefits are quite good. The API itself didn't change

Using `in` modifier passing `Mark` instances around as struct size exceeds pointer size.

Also introduced `ThrowHelper` to allow inlining `Mark` constructor.

## Benchmark Code

Uses the large YAML from issue https://github.com/aaubry/YamlDotNet/issues/519 . I noticed you don't have a benchmark project currently so I'm running my local one.

```csharp
[MemoryDiagnoser]
public class YamlStreamBenchmark
{
    private string yamlString = "";

    [GlobalSetup]
    public void Setup()
    {
        yamlString = File.ReadAllText(@"c:\work\saltern.yml");
    }

    [Benchmark]
    public void Load()
    {
        yamlStream.Load(new StringReader(yamlString));
    }
}
```

## Results

``` ini
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK=6.0.100
  [Host]     : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT
  DefaultJob : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT
```

### YamlStreamBenchmark

| **Diff**|Method|Mean|Error|Gen 0|Gen 1|Gen 2|Allocated|
|------- |-------|-------:|-------|-------:|-------:|-------:|-------:|
| Old |Load|96.71 ms|1.551 ms|5833.3333|2833.3333|1000.0000|81 MB|
| **New** |	| **73.13 ms (-24%)** | **1.319 ms** | **4428.5714 (-24%)** | **2000.0000 (-29%)** | **714.2857 (-29%)** | **63 MB (-22%)** |

### DIFF
--- a/YamlDotNet/Core/AnchorNotFoundException.cs
+++ b/YamlDotNet/Core/AnchorNotFoundException.cs
@@ -40,7 +40,7 @@ namespace YamlDotNet.Core
         /// <summary>
         /// Initializes a new instance of the <see cref="AnchorNotFoundException"/> class.
         /// </summary>
-        public AnchorNotFoundException(Mark start, Mark end, string message)
+        public AnchorNotFoundException(in Mark start, in Mark end, string message)
             : base(start, end, message)
         {
         }
--- a/YamlDotNet/Core/Events/DocumentStart.cs
+++ b/YamlDotNet/Core/Events/DocumentStart.cs
@@ -92,7 +92,7 @@ namespace YamlDotNet.Core.Events
         /// </summary>
         /// <param name="start">The start position of the event.</param>
         /// <param name="end">The end position of the event.</param>
-        public DocumentStart(Mark start, Mark end)
+        public DocumentStart(in Mark start, in Mark end)
             : this(null, null, true, start, end)
         {
         }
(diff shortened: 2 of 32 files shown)


aaubry:
> I noticed you don't have a benchmark project currently so I'm running my local one.

I used to have one to compare the performance across releases, but it was difficult to maintain and provided little value so it was removed. Feel free to add a benchmark if you feel it would be useful.

lahma:
@aaubry would you like it as part of this PR or separate?

aaubry:
> @aaubry would you like it as part of this PR or separate?


As you prefer. I was about to merge this, but I can wait.

lahma:
Please merge, I'll create a separate one! 😄

aaubry:

PR #649, opened 2021-11-15
Add YamlDotNet.Benchmark

lahma:
This adds a basic benchmark project and the saltern.yml (latest version which is now over 2MB) as GZIP compressed 73kb file.

Now uses VS 2022 image on appveyor to allow net6.0 target for benchmark. The benchmark project is bound by compatibility after all I guess.


lahma:
@aaubry here you go!

aaubry:

Issue #663, opened 2022-01-06
Serialized block comments with newline characters generates invalid output


krisrok:
**Describe the bug**
Using
```
[YamlMember(Description="Hello\nWorld)]
public int Foo { get; set; }
```

Expected result:
```
# Hello
# World
Foo: 0
```

Current implementation yields invalid YAML:
```
# Hello
World
Foo: 0
```

EdwardCooke:
The PR to fix this has been merged. Closing the issue.

Issue #662, opened 2022-01-06
Serialized block comments are not correctly indented right after a block starts


krisrok:
**Describe the bug**
Here's what's expected:
```
Person:
  # The person's name
  Name: Paul
  # The person's age
  Age: 50
```

The current implementation yields:
```
Person:
# The person's name
  Name: Paul
  # The person's age
  Age: 50
```

**To Reproduce**
Serialize any class with a reference type having a comment (using `YamlMemberAttribute`) on its first property.

EdwardCooke:
The PR fixing this was merged. Closing this.

PR #664, opened 2022-01-06
Fix/comment indents (and other comment-related issues)

krisrok:
This fixes incorrect comment indents (#662), see the updated `YamlCommentTests`.

It also fixes incorrect YAML output when using newline-characters inside of the `YamlMemberAttribute.Description` (#663): Now the Emitter simply writes multiple comment lines.

Also adds some support for flow mapping: Comments get stripped when inside a flow block.

#### Notes about inline comments
Also, as it is pretty related: I've added some code regarding writing inline comments, but they are not really supported as of now anyway: There's no easy to use public API.
In my branch (https://github.com/krisrok/YamlDotNet/tree/feature/serialize-inline-comments) there is some preliminary support for inline comments but I could not come up with a complete solution: Inline comments on non-emtpy sequences still land on the next line.


aaubry:

PR #674, opened 2022-02-08
Fix issue with long.MaxValue + added tests

### LINKED ISSUE #673: Deserialization of long.MinValue throws an OverflowException
**Describe the bug**
Using the latest version (11.2.1), deserializing `long.MinValue` is not possible. I just tried to update from `8.1.2` where it still worked.

**To Reproduce**
```
var deserializer = new Deserializer();
deserializer.Deserialize<long>(long.MinValue.ToString());
```

This throws the following exception:
```
YamlDotNet.Core.YamlException : (Line: 1, Col: 1, Idx: 0) - (Line: 1, Col: 21, Idx: 20): Exception during deserialization
  ----> System.OverflowException : Arithmetic operation resulted in an overflow.
```

This seems to be originating from `ScalarNodeDeserializer.DeserializeIntegerHelper`.

Falco20019:
Fixes #673 by using `unchecked` instead of `checked` cast. Also added test cases for avoiding regression.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -67,6 +67,47 @@ namespace YamlDotNet.Test.Serialization
             }
         }
 
+        public static IEnumerable<object[]> DeserializeScalarEdgeCases_TestCases
+        {
+            get
+            {
+                yield return new object[] { byte.MinValue, typeof(byte) };
+                yield return new object[] { byte.MaxValue, typeof(byte) };
+                yield return new object[] { short.MinValue, typeof(short) };
+                yield return new object[] { short.MaxValue, typeof(short) };
+                yield return new object[] { int.MinValue, typeof(int) };
+                yield return new object[] { int.MaxValue, typeof(int) };
+                yield return new object[] { long.MinValue, typeof(long) };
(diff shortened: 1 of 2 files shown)


EdwardCooke:
The issue I see with this is if someone gets a number greater than long.MaxValue (9223372036854775808) it will no longer be correct.

Example:

```csharp
Console.WriteLine(unchecked(-(long)9223372036854775809));
```

results in `9223372036854775807`

As such, I dont think we want to do this.

EdwardCooke:
One way to fix it would be to check if the number is the positive of the negative and if so, return long.minvalue.

```csharp
            if (isNegative)
            {
                long toCast;

                // the only real difference between long.minvalue and maxvalue is 1.
                // since the number part of long.minvalue is larger than long.maxvalue
                // we will check to see if it is the number and then return the min value
                // otherwise we will run it through checked to validate it is within our range
                if (result == 9223372036854775808)
                {
                    toCast = long.MinValue;
                }
                else
                {
                    toCast = checked(-(long)result);
                }

                return CastInteger(toCast, typeCode);
            }
            else
            {
                return CastInteger(result, typeCode);
            }
        }
```

EdwardCooke:
I've used parts of your tests but replaced the logic with what I posted above. Closing this in preference of #702.

PR #695, opened 2022-07-13
G17 was creating issues with doubles and floats

EdwardCooke:
Changes the number format to standard `G`. `G17` was causing rounding issues with doubles and floats. Numbers are still formatted correctly when needed for exponential notation, like `double.Epsilon`.


JuergenGutsch (approved):

aaubry:

PR #698, opened 2022-07-14
Allow set only properties during deserialization

### LINKED ISSUE #643: NullReferenceException when using setter-only property
**Describe the bug**
My datamodel contained a property with only purpose to transform value and save it to other property/field. Therefore, the property had only a setter defined. The library throws `NullReferenceException` (wrapped in `YamlException`) when attempting to deserialize into such model. Such exception tells nothing about what is happening and took me hours to figure it out. 

**To Reproduce**
Create simple class containing property with defined setter, but not getter. Try to deserialize yaml into the class.

EdwardCooke:
Currently we can't set set-only properties during deserialization. This allows set only properties to work.

Fixes #643


JuergenGutsch (approved):
great :-)

aaubry:

Issue #591, opened 2021-03-03
YamlScalarNode with "NULL" value should be written with surrounding quotes


AArnott:
When manually writing out a `YamlMappingNode` with `YamlScalarNode("NULL")` as a key or value, I get this:

```yaml
NULL: value
e: NULL
```

This is apparently a bad write job, given that `NULL` outside of quotes [actually means *null* to YAML](https://yaml.org/type/null.html), producing a null mapping node instead of `NULL` (the string) as a key in the map.

When I paste this into http://yaml-online-parser.appspot.com/, the json representation is (appropriately):

```json
{
  "null": "value", 
  "e": null
}
```

Note the change in case, because the YAML was read as *null* instead of as the intended *string* `NULL`. JSON has to surround a javascript `null` with quotes when it appears as a property name, but consider when the null is on the value side.

The emitted YAML should have been:

```yaml
"NULL": value
e: "NULL"
```

Note the quotes around the yaml-keyword, which then forces it to be read as a string literal by other YAML parsers:

```json
{
  "NULL": "value", 
  "e": "NULL"
}
```

## Repro steps

```cs
using System;
using System.IO;
using YamlDotNet.RepresentationModel;

var mappingNode = new YamlMappingNode(
    new YamlScalarNode("NULL"), new YamlScalarNode("value"),
    new YamlScalarNode("e"), new YamlScalarNode("NULL")
    );
var doc = new YamlDocument(mappingNode);
var stream = new YamlStream(doc);
var sb = new StringWriter();
stream.Save(sb);

Console.WriteLine(sb);
```

**Actual output**

```
&428684954 NULL: value
e: NULL
...
```

**Expected output**

```
&428684954 "NULL": value
e: "NULL"
...
```

## Deserialization fails

This bug becomes a more practical problem because the deserializer fails on null keys:

```cs
using System;
using System.Collections.Generic;
using System.IO;
using YamlDotNet.RepresentationModel;

var mappingNode = new YamlMappingNode(
    new YamlScalarNode("NULL"), new YamlScalarNode("V")
    );
var doc = new YamlDocument(mappingNode);
var stream = new YamlStream(doc);
var sb = new StringWriter();
stream.Save(sb);

Console.WriteLine(sb);

var deserializer = new YamlDotNet.Serialization.Deserializer();
var data = deserializer.Deserialize<Dictionary<string, string>>(new StringReader(sb.ToString()));

class Detail
{
    public string H { get; set; }
}
```

[...]

Issue #526, opened 2020-09-01
[Question] preserve property with empty string


srdjos:
Hi

as I cannot find the way, I would like to ask here. In my class I have string empty property:
name: ""

after serializing, the property is removed from .yml file. Also, I tried with the ChainedEventEmitter emitter, but the null or empty value does not come to the emitter. It looks like, the property is removed before the serializer hits the emitter:

Did I miss something in the serialization process?

```
public override void Emit(ScalarEventInfo eventInfo, IEmitter emitter)
        {
            if (eventInfo != null && eventInfo.Source.Type == typeof(string) && 
                (eventInfo.Source.Value == null || (string)eventInfo.Source.Value==""))
            {
                if (emitter != null)
                {
                    emitter.Emit(new Scalar(string.Empty));
                }
            }
            else
            {
                base.Emit(eventInfo, emitter);
            }
        }
```

EdwardCooke:
This will be fixed when the next release goes out.

PR #699, opened 2022-07-14
Allow quoting of specialized strings like null/true/false/numbers etc

### LINKED ISSUE #611: Null value not properly serialized
## Description
When serializing, null values (that is, null references, *not* strings with the value `"null"`) are written as an empty string instead of an actual YAML null. The possibility to serialize null values to actual null should be present either as a default or an option to the `SerializerBuilder` class.

As a side not, this issue is related to issues #591, #387 and #304 (ambiguous serialization of null-like values between YAML and CLR).

## To reproduce
```csharp
using System;
using System.Collections.Generic;
using System.IO;
using System.Text;
using System.Threading.Tasks;

class Program {
    static async Task Main(string[] args) {
        Dictionary<string, object> data = new() {
            ["null_test"] = null,
        };

        await using MemoryStream outStream = new();
        await using StreamWriter output = new(outStream, new UTF8Encoding());

        ISerializer serializer = new SerializerBuilder().DisableAliases().WithIndentedSequences().Build();
        serializer.Serialize(output, data);

        await output.FlushAsync();
        Console.WriteLine(new UTF8Encoding().GetString(outStream.ToArray()));
    }
}
```
### Expected output:
```yaml
null_test: null
````
or
```yaml
null_test: ~
```
### Actual output:
```yaml
null_test: 
```

EdwardCooke:
Title says it all :) It fixes a long running discussion in a non-breaking of existing behavior way. All tests passed as expected including new ones to account for the new functionality. I'm really excited about this one as it has been biting me for a long time with Kubernetes manifests.

Fixes
#611 
#591 
#526 
#387 

Uses parts of #532

### DIFF
--- a/YamlDotNet.Test/Core/EventsHelper.cs
+++ b/YamlDotNet.Test/Core/EventsHelper.cs
@@ -299,7 +299,7 @@ namespace YamlDotNet.Test.Core
 
             foreach (var property in expected.GetType().GetTypeInfo().GetProperties())
             {
-                if (property.PropertyType == typeof(Mark) || !property.CanRead)
+                if (property.PropertyType == typeof(Mark) || !property.CanRead || property.Name == "IsKey")
                 {
                     continue;
                 }
--- a/YamlDotNet.Test/Core/ScannerTests.cs
+++ b/YamlDotNet.Test/Core/ScannerTests.cs
@@ -438,7 +438,7 @@ namespace YamlDotNet.Test.Core
 
             foreach (var property in expected.GetType().GetTypeInfo().GetProperties())
             {
-                if (property.PropertyType != typeof(Mark) && property.CanRead)
+                if (property.PropertyType != typeof(Mark) && property.CanRead && property.Name != "IsKey")
                 {
                     var value = property.GetValue(actual, null);
                     var expectedValue = property.GetValue(expected, null);
--- a/YamlDotNet.Test/Serialization/DeserializerTest.cs
+++ b/YamlDotNet.Test/Serialization/DeserializerTest.cs
@@ -128,13 +128,28 @@ Value: bar
 True: null
 False: hello
 Null: true
+X:
 ";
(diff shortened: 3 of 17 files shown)


JuergenGutsch (approved):

aaubry:
I'm not sure that detecting that a scalar is a key should be made at the scanner level. In principle, the scanner's job is just to turn a stream of characters into a stream of tokens, without much semantics. Then the parser takes that stream and with knowledge of the grammar turns that stream of tokens into a stream of semantic events. In this context, I would expect the parser to be responsible for marking an event as a key.

Also, you are assuming that a key will always be a scalar, but YAML doesn't impose such restrictions. A key can be any node, including another mapping. For example, this is a mapping on which the first key is a scalar, the second is a mapping, and the third is a sequence:
```yaml
scalar: foo
{ a: mapping }: bar
[ a, sequence ]: baz
```

http://ben-kiki.org/ypaste/data/56801/index.html

aaubry:
I haven't tried, but since you added the `IsKey` property to the scalar token, you may have problems with anchors as well:
```yaml
a: &some_scalar this is also a key
*some_scalar : "will this key be handled correctly?"
```
http://ben-kiki.org/ypaste/data/56802/index.html

EdwardCooke:
@aaubry, those were some awesome YAML test cases. I added them to the great collection of test cases you already had created. The YAML you specified worked, along with adding numbers in to the mix, so it kept the datatypes the way I was expecting them to. The IsKey is probably named incorrectly, maybe something better would have been ParseDataType or something along those lines. Not sure on that though.

aaubry:
> @aaubry, those were some awesome YAML test cases. I added them to the great collection of test cases you already had created. The YAML you specified worked, along with adding numbers in to the mix, so it kept the datatypes the way I was expecting them to. The IsKey is probably named incorrectly, maybe something better would have been ParseDataType or something along those lines. Not sure on that though.

Good! I'm glad that this works 👍

aaubry:

PR #700, opened 2022-07-16
Block scalars can be keys too

### LINKED ISSUE #684: Deserialize fails for keys with newline
When using a complex key, that has newline char the serialization is successful but deserialization fails.

YAML example:
```yaml
? >-
  key

  a

  b
: >-
  value

  a

  b

```

Code:
```c#

var serializer = new SerializerBuilder().Build();

var data = new Dictionary<string, string>();
data.Add("key\na\nb", "value\na\nb");

var serialized = serializer.Serialize(data);

var deserialized = deserializer.Deserialize<Dictionary<string, string>>(serialized);

```

The version of the `YamlDotNet` package that I'm using is `11.2.1`.

Exception:

```
(Line: 1, Col: 1, Idx: 0) - (Line: 1, Col: 1, Idx: 0): Exception during deserialization -> Value cannot be null. (Parameter 'key')
```

EdwardCooke:
Fixes #684

### DIFF
--- a/YamlDotNet.Test/Serialization/DeserializerTest.cs
+++ b/YamlDotNet.Test/Serialization/DeserializerTest.cs
@@ -192,6 +192,30 @@ b: &number 1
             Assert.Equal("baz", item.Value);
         }
 
+        [Fact]
+        public void NewLinesInKeys()
+        {
+            var yaml = @"? >-
+  key
+
+  a
+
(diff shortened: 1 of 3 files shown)


JuergenGutsch (approved):

aaubry:

PR #701, opened 2022-07-16
fixup nuget package spec

JoC0de:
I was looking at some stuff and saw that there were some inconsistencies in the `YamlDotNet.nuspec`.
So here is a fix, so they are fixed before the next release.


aaubry:

PR #711, opened 2022-07-30
upgrade Bullseye from 3.6.0 to 4.0.0

adamralph:
I see you're using a few things from the `Bullseye.Internal` namespace—mainly `Palette` and `HostExtensions`. FYI I'm currently working on [making those public](https://github.com/adamralph/bullseye/pull/800) in Bullseye 4.1.0. One thing I'd like to confirm before releasing 4.1.0 is that the new API will remove the need for you to use the `Bullseye.Internal` namespace. But to do that, I think it's best if you've already moved to the Bullseye 4.0.0 API, so we can isolate the changes required for you to move from 4.0.0 to 4.1.0.

### DIFF
--- a/tools/build/BuildDefinition.cs
+++ b/tools/build/BuildDefinition.cs
@@ -93,7 +93,7 @@ namespace build
         {
             switch (options.Host)
             {
-                case Host.Appveyor:
+                case Host.AppVeyor:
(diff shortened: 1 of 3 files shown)


adamralph:
@aaubry any interest in this?

aaubry:
Sure, sorry I was unable to review this earlier. Thanks!

PR #720, opened 2022-09-13
Fix inconsistent line endings

DSchroer:
Update the line endings across the project to be consistent. 

# Motivation
Some tools complain when dealing with mixed line endings. The project currently has a mix of Windows `\r\n` and Unix style `\n`.


EdwardCooke:
I'm thinking a potentially better solution to this would be using `.gitattributes` to handle line endings. Maybe with the contents of something like this
```
*.sh EOL=lf
*.* EOL=crlf
```

I'm going off of memory here, so it may be incorrect values. But then, in theory, it'll be correct regardless of the platform (Linux aka Docker/Windows/etc) it is checked out on.

DSchroer:
Sure. Happy to add the adjusted `gitattributes`.

aaubry:
This change broke the build - https://ci.appveyor.com/project/aaubry/yamldotnet/builds/44921032

You need to be very careful with line endings because Linux doesn't expect `\r`. Personally I would have preferred not touching the line endings as everything was working correctly.

aaubry:
The the PR itself was building correctly - https://ci.appveyor.com/project/aaubry/yamldotnet/builds/44817924.
So this is probably a merge error.

aaubry:
The run.sh file hasn't been modified. I guess it's .gitattributes that is causing the problem. I'm not familiar with this file. I always configure git to never modify my line endings. @EdwardCooke do you have an idea of how to fix this?

EdwardCooke:
I'll get it fixed. Not sure why it would have failed either. gitattribes for .sh files was added a while ago.

EdwardCooke:
Got it fixed. *.* was overriding the *.sh.

EdwardCooke:
And when I just pushed up my changes to git attributes it pushed up the broken files, argh. Stupid gitattributes. Lesson learned there, don't do *.*.

aaubry:
This feature has been released in [version 12.0.2](https://github.com/aaubry/YamlDotNet/releases/tag/v12.0.2).

PR #721, opened 2022-09-13
Adjust null warning in ScalarNodeDeserializer

DSchroer:
Adjust `ScalarNodeDeserializer`  to allow nullable values in the deserializer in order to eliminate a warning. 

# Motivation

Modern C# nullables will emit warnings when nulls are not accounted for. The parse logic in `ScalarNodeDeserializer` is one case where those warnings are emitted.


aaubry:
This feature has been released in [version 12.0.2](https://github.com/aaubry/YamlDotNet/releases/tag/v12.0.2).

Issue #737, opened 2022-11-07
Serializer does not quote ambiguous strings


LaXiS96:
**Describe the bug**
Serializer does not quote strings in the source object containing `"Y"`, `"no"` and probably other ambiguous strings.
I am deserializing, modifying and then serializing docker-compose YAML files, which are picky when it comes to value types, like previously reported in #513.
The deserialization side works as expected, values which are strings both in the YAML file and in the POCO model are deserialized correctly as strings.
When serializing back though, those strings are not quoted and will be rejected by docker-compose, since they are interpreted as boolean values instead of strings, being valid boolean identifiers.

In my opinion the default behavior should be to quote those strings if the source object's value type is string, and allow opting out from this automatic quoting via a builder method.
In fact I tried using `WithQuotingNecessaryStrings()`, but there only difference was when the source string contained a valid number.

**To Reproduce**
Here is a quick DotNetFiddle: https://dotnetfiddle.net/TuZFTi

LaXiS96:
The quick workaround for scalar values is adding `[YamlMember(ScalarStyle = ScalarStyle.DoubleQuoted)]` to values that should always be quoted, but that workaround isn't available when working with dictionaries, like is the case with docker-compose service environment variables.

LaXiS96:
I have done some digging, it looks like alternative representations for boolean values were part of the YAML 1.1 specification (as documented [here](https://yaml.org/type/bool.html)) but were removed in 1.2.
The YAML parser used by docker-compose includes this passage in its [README](https://github.com/go-yaml/yaml/blob/v3.0.1/README.md?plain=1#L20):
```plain
YAML 1.1 bools (_yes/no, on/off_) are supported as long as they are being
decoded into a typed bool value. Otherwise they behave as a string. Booleans
in YAML 1.2 are _true/false_ only.
```

The easy fix would be to add the regex to [TypeAssigningEventEmitter](https://github.com/aaubry/YamlDotNet/blob/v12.0.2/YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs#L35) but I believe it would be a better idea to add a builder method to enable 1.1-compliant boolean parsing.
I might give it a try and open a pull request when I find the time to do it.

EdwardCooke:
This is actually released (the PR was re-created).

PR #738, opened 2022-11-07
Added option to quote YAML 1.1 boolean strings

LaXiS96:
This pull request addresses issue #737.
I added an optional parameter to `SerializerBuilder.WithQuotingNecessaryStrings` to opt into treating YAML 1.1 boolean strings like `"Yes"`, `"no"`, `"ON"`, `"Off"` as special strings that need to be quoted. Specification is [here](https://yaml.org/type/bool.html).
I also refactored the special strings regex to be compiled once at the construction of `TypeAssigningEventEmitter`.


perlpunk:
In my opinion such an option just for booleans doesn't make much sense. Not only booleans have changed between YAML 1.1 and 1.2, but also numbers. Have a look here for a compact comparison: https://perlpunk.github.io/yaml-test-schema/schemas.html
So you should take into account all the differences.

LaXiS96:
Thank you @perlpunk, I have updated the code to hopefully include all YAML 1.1 scalar styles.

EdwardCooke:
I’m going to review this in the next few days so we can get it moving through. In the meantime, you’ll need to rebase from master to fix the merge conflicts.

EdwardCooke:
The build hit the random bug during build. Can you push up another commit so it attempts a build again?

LaXiS96:
To be fair, I noticed all previous builds ended with the same `ResolveVersion` error, I haven't seen any build succeed.
Is there anything I can do on my side to ensure it builds?

EdwardCooke:
Instead of using master as your branch, try using a different branch name. It may be getting confused about having 2 master branches.

LaXiS96:
I renamed my main branch and had to open a new pull request (#767), which did build successfully.

PR #750, opened 2022-12-02
Throw an exception on duplicate keys

### LINKED ISSUE #536: Deserializing is not handling non-unique keys correctly
According to the YAML spec (and common sense) keys in YAML-mappings should be a set of unique values. I frequently encounter errors (originating from tiny typos of mine), where a document (or mapping within a document) like

```yml
fu: bar
fu: baz
```

is interpreted like 

```yml
fu: baz
```

and the duplicate key issue is silently ignored. 

Would it be possible to at least have an option to treat duplicate keys as deserialization errors?

nuttytree:
Deserializing YAML with non-unique keys throws a YamlException.  Resolves #536

### DIFF
--- a/YamlDotNet.Test/Serialization/DeserializerTest.cs
+++ b/YamlDotNet.Test/Serialization/DeserializerTest.cs
@@ -284,7 +284,25 @@ b: &number 1
         }
 
         [Fact]
-        public void Deserialize_YamlWithDuplicateKeys_ThrowsYamlException()
+        public void DeserializeWithDuplicateKeyChecking_YamlWithDuplicateKeys_ThrowsYamlException()
+        {
+            var yaml = @"
+name: Jack
+momentOfBirth: 1983-04-21T20:21:03.0041599Z
+name: Jake
+";
+
+            var sut = new DeserializerBuilder()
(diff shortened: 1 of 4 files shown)


EdwardCooke:
I would prefer if this was done as an option and defaulted to the current behavior. In the deserializer builder, we would add another method, something like `ThrowExceptionOnDuplicateKeys`. That would set the feature flag to enable or disable this code.

nuttytree:
I went without the option because duplicate keys are not valid YAML but I will update today to make it optional.

EdwardCooke:
Changing the behavior of the library may cause unexpected issues for the consumers which is why I would like to see it as optional to opt in. Maybe in a future major version release we’ll make it the default.
________________________________

nuttytree:
Sorry about the delay in getting this PR updated, been sucked into multiple things at work.  I have updated to make the duplicate key checking optional.  Let me know if you don't like the naming.

EdwardCooke:
Merged! Yay.

aaubry:
This feature has been released in [version 12.2.1](https://github.com/aaubry/YamlDotNet/releases/tag/v12.2.1).

Issue #753, opened 2022-12-06
Binaries using YamlDotNet built with the "Trim unused code" option fail at runtime


RobotsOnDrugs:
**Describe the bug**
Trimmed binaries fail to deserialize and throw an exception at runtime.

**To Reproduce**
Publish a project with the "Trim unused code" option.
Build log:
```
Build started...
1>------ Build started: Project: Project, Configuration: Release x64 ------
1>Project -> C:\Users\iamtheclaw\source\repos\Project\bin\x64\Release\net7.0-windows10.0.19041.0\Project.dll
2>------ Publish started: Project: Project, Configuration: Release x64 ------
2>Determining projects to restore...
2>Restored C:\Users\iamtheclaw\source\repos\Project\Project.csproj (in 829 ms).
[...]

Issue #740, opened 2022-11-10
support net 7's native aot


ivanjx:
**Is your feature request related to a problem? Please describe.**
i cant use this library with native aot on net 7 because this uses system.reflection.

**Describe the solution you'd like**
a solution similar to system.text.json is to use source generator.

**Describe alternatives you've considered**
excluding the types that i want to deserialize in `rd.xml` works.

**Additional context**
https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/

EdwardCooke:
When I get some time I’ll work on this. Not sure what it will take because we heavily use reflection, but supporting current frameworks and features is important. In the meantime, if you want to take a look at it we are happy to accept contributions.

EdwardCooke:
I’m still working on this. I got it to instantiate a class. It’s taking a lot of attributes on the methods so the classes and other objects don’t get stripped away. It also requires an attribute added to the class that gets deserialized so all of your yaml objects will require that as well.

ivanjx:
you might want to incorporate source generators.

https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/

https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/#what-is-the-system-text-json-source-generator

EdwardCooke:
Well that is an interesting idea @ivanjx . When I get some more time, I'll look in to them.

EdwardCooke:
Just to give you an update on this. I have a code generator/analyzer that will generate static factories for classes and property/field accessesors. It works for primitive types and anything marked with a `YamlSerializable` attribute. Generics do not work yet, so things like Dictionaries won't serialize/deserialize correctly.

I have not merged any of this in yet since it's not ready just yet.

I just wanted to let you know that I am making progress on getting this done. I am hoping to have this all done and ready by end of week.

EdwardCooke:
If you want to check out what I've got, here's my branch: https://github.com/EdwardCooke/YamlDotNet/tree/ec-aot

The demo app that shows how to implement it is in the `YamlDotNet.Core7AoTCompileTest` project. It's pretty simple.

EdwardCooke:
Pending a review of the PR I just made, we'll have the code in place in the main YamlDotNet library to support the AoT/Trimmed version of the library. Once approved and approved and merged in, I will push up a NuGet package containing the analyzer that you can try out to make sure it meets your needs.

EdwardCooke:
[...]

PR #758, opened 2022-12-14
Static Object Context for Ahead of Time compiling and Trimmed libraries

EdwardCooke:
Includes a source generator/analyzer that generate a static type inspector and object factory. These static objects allow the creation of objects and setting/getting of object properties and fields so YamlDotNet serializer will work with trimmed libraries and ahead of time compilation because it removes the use of reflection when these static objects are used during serialization and deserialization.

Currently it has the following requirements, support and limitations:

* All objects in the YAML object graph must be annotated with the `YamlSerializable` attribute
* `YamlIgnore` and `YamlMember` attributes are supported and respected
* `Dictionary<,>` and `List<>` property/field types are supported.
* There is a helper method on the `SerializerBuilder` and `DeserializerBuilder` that will configure the serializer and deserializer to not use the reflection based objects and instead use the static ones.
* Arrays are supported
* Type overrides are not tested, if you annotate the real class with `YamlSerializable` it might work

This PR will not create or push an analyzer package to NuGet. That will come later.

This is to support creating an analyzer package manually that can be tested by the users in #740 and #753.

@aaubry , I know you are trying to take a more passive role in YamlDotNet, but I would really appreciate a review of this and the subsequent PR, at a potentially near future date, where we push up the official analyzer so people can use it for real.


aaubry:
This feature has been released in [version 12.3.0](https://github.com/aaubry/YamlDotNet/releases/tag/v12.3.0).

PR #759, opened 2022-12-15
Allow braces in non-flows

### LINKED ISSUE #757: Values starting with `-` but otherwise being string are treated like a key start
**Describe the bug**
Please just see the example below as it tells everything.

**To Reproduce**
```csharp

var input = @"
Value: -[123]
";

var deserializer = new Deserializer();
deserializer.Deserialize(new StringReader(input));
```
Running this leads to an exception (Invalid key indicator format), while this should be simply deserialized to a string. It worked properly on YamlDotNet 8.0.0.

EdwardCooke:
Fixes #757

### DIFF
--- a/YamlDotNet.Test/Core/ScannerTests.cs
+++ b/YamlDotNet.Test/Core/ScannerTests.cs
@@ -414,6 +414,35 @@ namespace YamlDotNet.Test.Core
             }
         }
 
+        [Fact]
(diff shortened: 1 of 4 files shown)


aaubry:
This feature has been released in [version 12.3.0](https://github.com/aaubry/YamlDotNet/releases/tag/v12.3.0).

PR #767, opened 2023-01-14
Option to quote YAML 1.1 strings

LaXiS96:
See previous PR #738


YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs
```diff
@@ -32,28 +32,54 @@ public sealed class TypeAssigningEventEmitter : ChainedEventEmitter
...
-            @"^("
-                 + @"null|Null|NULL|\~"
-                 + @"|true|True|TRUE|false|False|FALSE"
-                 + @"|[-+]?[0-9]+|0o[0-7]+"
-                 + @"|0x[0-9a-fA-F]+"
-                 + @"|[-+]?(\.[0-9]+|[0-9]+(\.[0-9]*)?)([eE][-+]?[0-9]+)?"
-                 + @"|[-+]?(\.inf|\.Inf|\.INF)"
-                 + @"|\.nan|\.NaN|\.NAN"
+        private readonly Regex isSpecialStringValue_Regex;
+        private static readonly string SpecialStrings_Pattern = ""
```
EdwardCooke, line 36: You don't need the `""` at the end of this line. And then remove the `+` at the beginning of the next.

YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs
```diff
@@ -32,28 +32,54 @@ public sealed class TypeAssigningEventEmitter : ChainedEventEmitter
...
+                + @"|[-+]?(0|[1-9][0-9_]*)" // int base 10
+                + @"|[-+]?0x[0-9a-fA-F_]+" // int base 16
+                + @"|[-+]?[1-9][0-9_]*(:[0-5]?[0-9])+" // int base 60
+                + @"|[-+]?([0-9][0-9_]*)?\.[0-9_]*([eE][-+][0-9]+)?" // float base 10
+                + @"|[-+]?[0-9][0-9_]*(:[0-5]?[0-9])+\.[0-9_]*" // float base 60
+                + @"|[-+]?\.(inf|Inf|INF)"
+                + @"|\.(nan|NaN|NAN)"
+            + @")$";
 
-        public TypeAssigningEventEmitter(IEventEmitter nextEmitter, bool requireTagWhenStaticAndActualTypesAreDifferent, IDictionary<Type, TagName> tagMappings)
```
EdwardCooke, line 52: Removing this constructor becomes a breaking change, which we avoid unless its absolutely necessary. Can you refactor this so it doesn't remove it? Instead, your constructor should call the one that was above this, where it expects the quoted string.

YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs
```diff
@@ -32,28 +32,54 @@ public sealed class TypeAssigningEventEmitter : ChainedEventEmitter
...
+                + @"|true|True|TRUE|false|False|FALSE"
+                + @"|[-+]?[0-9]+" // int base 10
+                + @"|0o[0-7]+" // int base 8
+                + @"|0x[0-9a-fA-F]+" // int base 16
+                + @"|[-+]?(\.[0-9]+|[0-9]+(\.[0-9]*)?)([eE][-+]?[0-9]+)?" // float number
+                + @"|[-+]?(\.inf|\.Inf|\.INF)"
+                + @"|\.nan|\.NaN|\.NAN"
             + @")$";
 
-        public TypeAssigningEventEmitter(IEventEmitter nextEmitter, bool requireTagWhenStaticAndActualTypesAreDifferent, IDictionary<Type, TagName> tagMappings, bool quoteNecessaryStrings)
```
EdwardCooke, line 46: Removing this is a breaking change. You should call this constructor from the new one you are creating.

LaXiS96:
I applied your suggestions 👍

PR #768, opened 2023-01-15
Fixed DateTime serialization with JSON using single quotes

### LINKED ISSUE #587: DateTime JSON compatible serialization doesn't put the date string in double quotes
It seems that **JsonEventEmitter** doesn't set the `eventInfo.Style` correctly for `DateTime`, resulting in the date/time string ending up in single quotes what is not JSON compatible. For example:
  ```
using System;
  
  namespace YamlDotNetDateTimeBugReport
  {
      class TestData
      {
          public DateTime Date => DateTime.Now;
      }
  
      class Program
      {
          static void Main(string[] args)
          {
              ISerializer jsonSerializer = new SerializerBuilder()
                  .JsonCompatible()
                  .Build();
              Console.WriteLine(jsonSerializer.Serialize(new TestData()));
          }
      }
  }

````

results in 

`{"Date": '2021-02-24T10:20:12.3763590+01:00'}`

The fix is simple, just adding
`eventInfo.Style = ScalarStyle.DoubleQuoted;`
under `case TypeCode.DateTime` in **JsonEventEmitter.cs**

### PR DESCRIPTION
**Pull request by EdwardCooke**
Fixes #587 

When using `JsonCompatible` we will now register a `DateTimeConverter`.

If the user is adding a `DateTimeConverter` object before calling `JsonCompatible` on the serializer, this call will fail. If the user is adding some other, custom, `DateTime` converter, the one added by `JsonCompatible` will take priority, and the users converter will be skipped.

If the user wants to specify their own `DateTimeConverter` or some other `DateTime` converter object, they will need to do it after calling `JsonCompatible`.

If they are passing their own `DateTimeConverter`, they can now specify the optional constructor argument, `doubleQuotes`, so it will dish out valid JSON.

### DIFF
--- a/YamlDotNet.Test/Serialization/DateTimeConverterTests.cs
+++ b/YamlDotNet.Test/Serialization/DateTimeConverterTests.cs
@@ -564,6 +564,16 @@ namespace YamlDotNet.Test.Serialization
 
             serialised.Should().ContainEquivalentOf($"datetime: {formatted}");
         }
+
+        [Fact]
+        public void JsonCompatible_EncaseDateTimesInDoubleQuotes()
+        {
+            var serializer = new SerializerBuilder().JsonCompatible().Build();
+            var testObject = new TestObject { DateTime = new DateTime(2023, 01, 14, 0, 1, 2) };
+            var actual = serializer.Serialize(testObject);
+
+            actual.TrimNewLines().Should().ContainEquivalentOf("{\"DateTime\": \"01/14/2023 00:01:02\"}");
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by stan-sz** [comment]
@EdwardCooke - my local test runs fail with

```
[xUnit.net 00:00:02.63]     JsonCompatible_EncaseDateTimesInDoubleQuotes [FAIL]
  Failed JsonCompatible_EncaseDateTimesInDoubleQuotes [7 ms]
  Error Message:
   Expected string to contain equivalent of "{\"DateTime\": \"01/14/2023 00:01:02\"}" but found "{\"DateTime\": \"01/13/2023 23:01:02\"}"
```

My guess is that the constructor in the test case is missing the `DateTimeKind` as:

```csharp
var testObject = new TestObject { DateTime = new DateTime(2023, 01, 14, 0, 1, 2, DateTimeKind.Utc) };
```

EdwardCooke:
This flaky test will be fixed in the next release.

PR #771, opened 2023-01-20
Follow up on duplicate key checking

stan-sz:
Follow up on #536 and extend the functionality to `DictionaryNodeDeserializer`.

### DIFF
--- a/YamlDotNet/Serialization/NodeDeserializers/ObjectNodeDeserializer.cs
+++ b/YamlDotNet/Serialization/NodeDeserializers/ObjectNodeDeserializer.cs
@@ -55,17 +55,16 @@ namespace YamlDotNet.Serialization.NodeDeserializers
             var implementationType = Nullable.GetUnderlyingType(expectedType) ?? expectedType;
 
             value = objectFactory.Create(implementationType);
-            var consumedProperties = new List<string>();
+            var consumedProperties = new HashSet<string>(StringComparer.Ordinal);
             while (!parser.TryConsume<MappingEnd>(out var _))
             {
                 var propertyName = parser.Consume<Scalar>();
-                if (duplicateKeyChecking && consumedProperties.Contains(propertyName.Value))
+                if (duplicateKeyChecking && !consumedProperties.Add(propertyName.Value))
                 {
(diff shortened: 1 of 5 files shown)


YamlDotNet/Serialization/NodeDeserializers/ObjectNodeDeserializer.cs
```diff
@@ -55,17 +55,16 @@ public bool Deserialize(IParser parser, Type expectedType, Func<IParser, Type, o
             var implementationType = Nullable.GetUnderlyingType(expectedType) ?? expectedType;
 
             value = objectFactory.Create(implementationType);
-            var consumedProperties = new List<string>();
+            var consumedProperties = new HashSet<string>(StringComparer.Ordinal);
             while (!parser.TryConsume<MappingEnd>(out var _))
             {
                 var propertyName = parser.Consume<Scalar>();
-                if (duplicateKeyChecking && consumedProperties.Contains(propertyName.Value))
+                if (duplicateKeyChecking && !consumedProperties.Add(propertyName.Value))
```
stan-sz, line 62: this approach lets the hashset grow only if the check is enabled

EdwardCooke:
I’m hoping to have time today to look at this and get it merged in

YamlDotNet/Serialization/NodeDeserializers/StaticDictionaryNodeDeserializer.cs
```diff
@@ -29,8 +29,8 @@ public class StaticDictionaryNodeDeserializer : DictionaryNodeDeserializer
     {
         private readonly ObjectFactories.StaticObjectFactory _objectFactory;
 
-        public StaticDictionaryNodeDeserializer(ObjectFactories.StaticObjectFactory objectFactory)
```
EdwardCooke, line 32: Since this is part of the breaking changes for static serializer/deserializer I'm not going to have you fix this, but, to make this change in a non-breaking way, you would call your new constructor with these arguments and set the duplicateKeyChecking to false.

Example:
```csharp
public StaticDictionaryNodeDeserializer(ObjectFactories.StaticObjectFactory objectFactory)
    : this(objectFactory, false)
```

Issue #731, opened 2022-10-14
Deserializing subclasses of abstact class


Ficus-x:
Hi! I have an abstract class and its sub classes. In list of that abstact class i add its subclasses and it serializes perfectly. But when i want to deserialize it, it throws error `No node deserializer was able to deserialize the node into...`. Can i somehow avoid it?
My deserializer looks like:
```public static IDeserializer Deserializer { get; set; } = new DeserializerBuilder()
            .WithTypeConverter(new VectorsConverter())
            .WithTypeConverter(new ColorConverter())
            .WithTypeConverter(new AttachmentIdentifiersConverter())
            .WithNamingConvention(UnderscoredNamingConvention.Instance)
            .WithNodeDeserializer(inner => new ValidatingNodeDeserializer(inner), deserializer => deserializer.InsteadOf<ObjectNodeDeserializer>())
            .IgnoreFields()
            .IgnoreUnmatchedProperties()
            .Build();

**Comment by edwardcookemacu** [comment]
Let me rephrase this so I can make sure I understand your question you have. You have an abstract class, `Foo`. With 2 implementing classes, `Bar` and `Baz`. You want YamlDotNet to automatically deserialize into `Bar` or `Baz`? In the example classes bellow you would be trying to deserialize Main.

I suspect it's throwing an error when trying to instantiate the abstract class, since that is what is probably the object type of your property.


```csharp
public abstract class Foo {
 public string X
}
public string Main {
 public Foo Prop { get; set; }
}

public class Bar : Foo {
 public string Prop1 { get; set; }
}

public class Baz : Foo {
 public string Prop2 { get; set; }
}
```

edwardcookemacu:
If that is the case, then the solution in this issue may help you.

https://github.com/aaubry/YamlDotNet/issues/605#issuecomment-817424757

EdwardCooke:
No response for a few months, closing the issue.

Issue #605, opened 2021-04-11
Dynamic type resolution during deserialization based on serialized content


nitz:
## Is your feature request related to a problem? Please describe.
This one is a bit of a doozy, and touches on several issues I've found are already open or related. To tag them quickly: The workflow I'm looking to support is related somewhat to the feature added in #601, and partially described or would be affected by the topics in #408, #459, and maybe tangent related to  #529 and #593.

Before I get started here as well: I am perfectly ready to be told "you dummy, it's super easy and you're just missing it!" So please do point out if I'm just glazing over features I shouldn't have!

Essentially, I've got some YAML I don't control. Nodes are untagged, but differ slightly in what they may contain. Other applications using this YAML react to each node by handling the node as a specific type based on the appearance of specific keys in direct children of the node. This is my goal as well.

With (my understanding) of the numerous modifications I can make to the parse/deserialization process, it seems I'm very close to being able to react the way I'd like, but each method I've tried to solve it has had some various shortcomings ranging from not able to have enough information to construct the proper type, to probably being able to handle it but am left parsing out the entire object on my own. I've got a more concrete example in the section below.

## Example problem

I apologize if this gets long winded or too verbose, just want to try and paint a complete picture. Please excuse any syntax errors, as I'm just typing this up all freeform! 🙂

I'll start with two example documents, and the data classes used to represent them.

First, my list of foos.

```yaml
# My Foos
name: "Example A"
url: "http://example.com/a"
timeout: 30
max_tries: 3
# other properties follow...
---
name: "Example B"
url: "http://example.com/b"
# other properties follow...
---
name: "Example C"
ftp: "192.0.2.14"
# notice no `url` key, instead `ftp`.
# different properties follow...
```

Second, my foo settings (and other settings) file.

```yaml
# My Foo Settings, each key under defaults would be 
# a specific concrete foo to hold all the related data to their types.
defaults:
  all:
    timeout: 30
  url:
    max_tries: 1
  ftp:
    timeout: 10
  # other keys here.
verbose: true
# other various settings.
```

And now, the .NET data model:

```csharp
// I'd like to keep objects as immutable, but that can be a future problem ;)
interface IFoo 
{
    string Name { get; set; }
    int Timeout { get; set; }
}

class FooBase : IFoo
{
    /* assume public virtual IFoo impl. */
}

class UrlFoo : FooBase
{
    public virtual string Url { get; set; }
    public virtual int MaxTries { get; set; }
    /* other properties as needed */
[...]

Issue #343, opened 2018-08-19
Determining the type from a field


felixfbecker:
I want do deserialize Kubernetes YAML that looks like this:

```yml
apiVersion: v1
kind: Deployment
metadata:
  name: proxy
  namespace: prod
  # ...
spec:
  minReadySeconds: 10
  replicas: 1
  revisionHistoryLimit: 10
  # ...
```
```yml
apiVersion: v1
kind: Service
metadata:
  name: proxy
  namespace: prod
  # ...
spec:
  ports:
  - name: http
    port: 8080
    targetPort: http
  type: ClusterIP
  # ...
```

The type of the object (what other fields besides `apiVersion` and `kind` the object will have) is determined by the `kind` field. Depending on the value of `kind`, I need to map this YAML to e.g. either a `Service` or a `Deployment` model class.

I tried TypeResolvers and TypeConverters but couldn't figure it out. Is this possible somehow with YamlDotNet?

bloudraak:
I have a similar need when dealing with polymorphic objects. Is there an example for `TypeResolvers` and `TypeConverters`?

gliljas:
I also have a similar need. My current workaround is a two phase serialization. I first deserialize to Dictionary<object,object>, then traverse the key value pairs and when I have determined the type (kind), I serialize the value and deserialize it using the determined type. It's ugly, but it works.

thisisthedave:
I've implemented variations on this problem with a slightly different 2-phase deserialization. Here's three different ways to handle similar situations.

**Polymorphic Document Type**
Use 2-phase deserialization. This is like the kubernetes example @felixfbecker posted. Instead of deserializing to a `Dictionary`, deserialize to a custom type with just a `kind` field:
```
    public class Kind
    {
        public string Kind { get; set; }
    }

    public void DoImportantStuff()
    {
        Deserializer deserializer = ...;
        Kind k = deserializer.Deserialize<Kind>(input);
        if (k.Kind == "Deployment")
            Deployment d = deserializer.Deserialize<Deployment>(input);
        // etc...
    }
```

**Use a Tag to Indicate the Concrete Type**
When deserializing polymorphic fields inside the object hierarchy, I indicate the concrete type with a yaml tag. You can do this either by using `DeserializerBuilder.WithTagMapping` (easier) or by implementing a custom `INodeTypeResolver` (more flexible).

Yaml
```
target: !TargetingData # !TargetingData indicates the concrete type of the field
[...]

Issue #36, opened 2013-06-28
Question about generic serialization


bj0:
Here's a contrived example to demonstrate my question:

Say I have some messages defind by their payloads, and a generic envelope
that I stick a payload in and send/receive with.  This is generally how I would
structure that classes:

~~~{.cs}
class RequestMessage
{
    [YamlAlias('request')]
    string Request { get; set }
    
    [YamlAlias('requester_name')]
    string RequesterName { get; set; }
}

class ConfigMessage
{
    [YamlAlias('a')]
    string SettingA { get; set; }
    
    [YamlAlias('b')]
    int SettingB { get; set; }
}


class MessageEnvelope<TPayload>
{
    [YamlAlias('message_type')]
    string Type { get; set; }
    
    [YamlAlias('payload')]
    TPayload Payload { get; set; }
}
~~~

One issue I've been trying to figure out is how do you make this work well
with serialization, or more accurrately deserialization.  If I create and object
and serialize it like so:

~~~{.cs}
var ys = new Serializer();
var sr = new StringWriter();

var msg = new MessageEnvelope<ConfigMessage>(){
        Type="config", 
        Payload = new ConfigMessage(){
            SettingA = "red",
            SettingB = 5
            }
        };

ys.Serialize(sr, msg);

sr.ToString().Dump();
~~~

Ideally it creates yaml documents that looks like:

~~~{.yaml}
---
message_type: request
payload:
    request: alpha
    requester_name: bob
...

---    
message_type: config
payload:
    a: red
    b: 5
...
~~~

Which it does now (I just tested, it use to just give 'payload: {}', but must 
have been fixed in a recent update). 

The question is, how do you deserialize this in one pass to an object, when you
don't know what the payload type is before deserializing?

Currently I'm treating the payload as a separate serialized object, which yields
yaml like this:

~~~{.yaml}
---
message_type: request
payload: 'request: alpha\nrequester_name: bob\n\n'
...

---
message_type: config
payload: 'a: red\nb: 5\n\n'
...
~~~

This isn't nearly as clean and readible, and I was hoping for a good way to 
do this, does anyone have any ideas?

jmo-gelsight:
This is marked as "completed".  May I ask, how was this completed?  Because I am facing the same issue.

EdwardCooke:
[...]

PR #774, opened 2023-02-02
Buffered deserialisation

JJ11teen:
Hello!

This is a draft PR to add support for deserialising yaml into different types, depending on values found during deserialisation. Specifically, it addresses scenarios mentioned in: #36 #343, #605, #731.

The foundational code included in this PR was written by @atruskie in a [gist available here]( https://gist.github.com/atruskie/bfb7e9ee3df954a29cbc17bdf12405f9). I am opening this PR with their permission.

Notably, this PR only includes a subset of the functionality, specifically the ability to discern type by looking for a key that has the desired type encoded in its value. This PR does not include the functionality to discern type by looking at the presence of unique keys. It does lay the foundation to bring in this additional functionality in the future if desired.

My use case is very similar to the kubernetes use case (where there is a "kind" key), and this PR should satisfy the kubernetes use case.

In addition to less functionality, I have also:
* Removed specific use case types that @atruskie included as examples in the gist.
* Added functionality to specify a max depth/length for buffering values during deserialisation.
* Added a helper method to DeserializerBuilder.

Here is an example of usage:
```csharp
    .WithBufferedNodeDeserializer(options => {
        options.AddKeyValueDiscriminator<UserSpecifiedBaseClass>(
            "kind",
            ("deployment", typeof(UserSpecifiedSubclassA)),
            ("service", typeof(UserSpecifiedSubclassB)));
    })
    .Build();
```

I wanted to put this up and gauge interest before working on this more, but the remaining things I know I need to do before merging:
* Comments/code style changes - I have yet to read up on coding conventions but will do
* Tests - I have only tested on net7 currently
* Anything else?

Some other things that could be done in this or future PRs:
* Add back in the other use of buffered deserialisation from @atruskie's gist?
* Handle various case-translations for keys/values?

It is also worth mentioning @EdwardCooke that this also works with the AOT branch we've been testing, it just needs a similar helper method on the `StaticDeserializerBuilder` as that file is not in the codebase yet.

Let me know if this is desired and I'll keep working on it, and thanks for maintaining this library (:


atruskie:
Love this; let me know if you want some help to finish it off.

I'd say that the resolver you didn't implement is equally useful and is probably worth adding. I found it more common that yaml files did not include a discriminator key. 

Or if you don't agree, I think it would be better to provide a configuration option method to add a custom ITypeDiscriminator so others can extend as they like.

EdwardCooke:
This is great! I've seen a few questions/issues related to this. It'll be nice to have it built in.

A few things I see right off the top that would be great to have addressed
* Comments on public classes and methods
* Adding a way to add a custom `IValueTypeDescriminator` would be a good thing and allow for additional flexibility
* It may be beneficial to mark your methods on your `KeyValueTypeDiscriminator` as virtual so people can use that class and extend it easily.
* It looks like the determined class has to be implementing a base class, and only the inherited class is checked. That may not always be the case. Example use cases where that would not be sufficient.
    * The property being assigned may be an `object` type, not requiring it to be abstract or doing any kind of inheritance.
    * It may also be the class itself, not abstract or inherited.
    * Deep inheritance (class c inherits b, which inherits a and the property type is a).
    * Implementing an interface and the property type is that interface
    * One possible solution would be to have the discriminator do the check to see if it can assign itself to the type, adds more flexibility. And in the default implementation it would call the isassignablefrom method on the type. Pretty sure that's what it is.
 
For the tests, add them to the tests project. It runs them against net47, net 3.1 (for netstandard2.0), net6.0 and net7.0.

JJ11teen:
Thanks both, good to know there is appetite for this! 😊 I'll work on the things mentioned and let you know when its ready for a closer look (:

JJ11teen:
Okay, I think I have addressed what I consider "design" changes at this point. Still have comments & tests to do though (:
Changes:
* Renamed "IValueTypeDiscriminator" -> "ITypeDiscriminator" and moved into a sub namespace/directory
* Added in @atruskie's other "UniqueKeyTypeDiscriminator" functionality
* Pulled out "TryFindMappingEntry" into existing "ParserExtensions" file
* Added "AddTypeDiscriminator" to Options class to enable users to add their own custom ITypeDiscriminator implementation more easily
* Updated "BufferedNodeDeserializer" to use "IsAssignableFrom", and have removed null & identity type checks - but have kept the concept of a "BaseType" on all discriminators. I believe this supports the scenarios you described @EdwardCooke:
  * The property being assigned may be an object type, not requiring it to be abstract or doing any kind of inheritance.
  * It may also be the class itself, not abstract or inherited.
  * Deep inheritance (class c inherits b, which inherits a and the property type is a).
  * Implementing an interface and the property type is that interface

* However I stopped short of having the discriminator doing the assignment check, as my understanding is that this would result in everything (ie the whole yaml stream) being buffered. The current implementation only starts buffering when the expected type has an associated discriminator. Users can still explicitly use object for the BaseType as you describe, but it is the users responsibility. We can make the BaseType default to object if you think this is a common enough use case and always buffering by default is worth it?

JJ11teen:
Actually the BaseType = object scenario is more complex than I previously thought - won't get to it today but will keep thinking (:

JJ11teen:
Okay, I have reworked things such that BaseType = object is now supported (see tests).

In order to accomplish this, I have made the `BufferedNodeDeserializer` now wrap the entire list of NodeDeserializers that the Builder has specified. I found this was needed as otherwise the `DictionaryNodeDeserializer` would take precedence when BaseType=object, and the `BufferedNodeDeserializer` was never reached. I also tried just inserting the previous implementation earlier in the builder's deserializer list, but that failed when the deserialized type had a Dictionary as a sub type (as in the case for kubernetes labels/annotations). I'm wondering if you're aware of a better way to handle this NodeDeserializer dependency than what I've got now @EdwardCooke ?

JJ11teen:
I can't think of a better solution without changing the signature for `INodeDeserializer`.
```csharp
public interface INodeDeserializer
{
    bool Deserialize(IParser reader, Type expectedType, Func<IParser, Type, object?> nestedObjectDeserializer, out object? value);
}
```

The only output of each `INodeDeserializer` is the deserialized value, (and whether or not the deserializer was able to deserialize any value). This is at odds with the desire for the BufferedNodeDeserializer, which is to "Buffer and deserialize to determine a target type, then reset the buffer and redirect to an appropriate existing INodeDeserializer that matches the newly determined target type".

This is why the BufferedNodeDeserializer needs to wrap all other INodeDeserializers. However all the others also need to exist outside of the BufferedNodeDeserializer for cases when buffering isn't needed/desired.

I don't like the duplication of all of the INodeDeserializers but I think it is a trade off between that and more significantly impacting the existing codebase/functionality/user expectations. Does anyone have any ideas for a more elegant solution? 😂

JJ11teen:
Hey @EdwardCooke, just a ping that this is ready for review (:
I rebased on master and added the helper to the static builder yesterday too.

YamlDotNet/Core/ParserExtensions.cs
```diff
@@ -147,5 +147,50 @@ public static void SkipThisAndNestedEvents(this IParser parser)
         {
             return Accept<T>(parser, out var _);
         }
+
+        public static bool TryFindMappingEntry(this IParser parser, Func<Scalar, bool> selector, out Scalar? key, out ParsingEvent? value)
```
EdwardCooke, line 151: Can you add some xml comments on here to describe what this method is/does?

YamlDotNet/Core/ParserExtensions.cs
```diff
@@ -147,5 +147,50 @@ public static void SkipThisAndNestedEvents(this IParser parser)
...
+        {
+            parser.Consume<MappingStart>();
+            do
```
EdwardCooke, line 154: Should this be a `while (parser.Current != null)` to avoid a null reference on your `switch` statement? Not sure if it could be null or if it would break, but it might.

YamlDotNet/Serialization/BufferedDeserialization/ParserBuffer.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.Collections.Generic;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+
+namespace YamlDotNet.Serialization.BufferedDeserialization
+{
+        public class ParserBuffer : IParser
```
EdwardCooke, line 8: Can you fix the endentation and add some xml comments

YamlDotNet/Serialization/BufferedDeserialization/ParserBuffer.cs
```diff
@@ -0,0 +1,51 @@
...
+                if (maxLength > -1 && buffer.Count > maxLength)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(parserToBuffer), "Parser buffer exceeded max length");
+                }
+            } while (depth >= 0);
+
+            current = buffer.First;
+        }
+
+        public ParsingEvent? Current => current?.Value;
```
EdwardCooke, line 38: Can you add xml comments on all of the public members?

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/IValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using YamlDotNet.Core;
+
+namespace YamlDotNet.Serialization.BufferedDeserialization.TypeDiscriminators
+{
+    public interface ITypeDiscriminator
```
EdwardCooke, line 6: Can you put some xml comments on this?

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/KeyValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public class KeyValueTypeDiscriminator : ITypeDiscriminator
+    {
+        public Type BaseType { get; private set; }
+        private readonly string targetKey;
+        private readonly IDictionary<string, Type> typeMapping;
+
+        public KeyValueTypeDiscriminator(Type baseType, string targetKey, IDictionary<string, Type> typeMapping)
+        {
+
```
EdwardCooke, line 18: It's nit picky, but could get rid of the extra whitespace?

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/KeyValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,67 @@
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+using YamlDotNet.Serialization;
+
+namespace YamlDotNet.Serialization.BufferedDeserialization.TypeDiscriminators
+{
+    public class KeyValueTypeDiscriminator : ITypeDiscriminator
```
EdwardCooke, line 10: Can you add some xml comments on this class and its public members?

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/KeyValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                // read the value of the discriminator key
+                if (value is Scalar valueScalar)
+                {
+                    suggestedType = CheckName(valueScalar.Value);
+                    return true;
+                }
+                else
+                {
+                    throw new Exception($"Could not determine {BaseType} to deserialize to, {targetKey} has an empty value");
```
EdwardCooke, line 46: Should this case just be ignored? When would it fall in this else statement get used? Throwing an exception here seems like it could be dangerous.

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/KeyValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,67 @@
...
+        private Type CheckName(string value)
+        {
+            if (typeMapping.TryGetValue(value, out var childType))
+            {
+                return childType;
+            }
+
+            var known = string.Join(",", typeMapping.Keys.ToArray());
+            
+            throw new Exception($"Could not determine {BaseType} to deserialize to, expecting '{targetKey}' to be one of: {known}, but got '{value}'");
```
EdwardCooke, line 64: If instead of throwing an exception, you returned null, here and false above, would you be able to have multiple discriminators triggering on different values of the key?

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/UniqueKeyTypeDiscriminator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private readonly IDictionary<string, Type> typeMapping;
+
+        public UniqueKeyTypeDiscriminator(Type baseType, IDictionary<string, Type> typeMapping)
+        {
+            foreach (var keyValuePair in typeMapping)
+            {
+                if (!baseType.IsAssignableFrom(keyValuePair.Value))
+                {
+                    throw new ArgumentOutOfRangeException($"{nameof(typeMapping)} dictionary contains type {keyValuePair.Value} which is not a assignable to {baseType}");
```
EdwardCooke, line 22: Can you add `nameof(typeMapping)` to the arguments on this exception? That way it's automatically included in this exception.

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/KeyValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,67 @@
...
+        private readonly IDictionary<string, Type> typeMapping;
+
+        public KeyValueTypeDiscriminator(Type baseType, string targetKey, IDictionary<string, Type> typeMapping)
+        {
+            
+            foreach (var keyValuePair in typeMapping)
+            {
+                if (!baseType.IsAssignableFrom(keyValuePair.Value))
+                {
+                    throw new ArgumentOutOfRangeException($"{nameof(typeMapping)} dictionary contains type {keyValuePair.Value} which is not a assignable to {baseType}");
```
EdwardCooke, line 23: Can you add nameof(typeMapping) to the arguments on this exception? That way it's automatically included in this exception.

JJ11teen:
Thanks for the review @EdwardCooke! Will aim to get back to this over the coming week 😊

YamlDotNet/Core/ParserExtensions.cs
JJ11teen: Makes sense to me (and I've made the change and tests pass), but I'll admit this function is not really one I fully understand. @atruskie if you have time could you double check this logic change and the xml comment I've added?

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/KeyValueTypeDiscriminator.cs
JJ11teen: Yep, makes sense to me to make the discriminator more passive and just ignore things it doesn't understand (:

JJ11teen:
I think I've addressed everything @EdwardCooke! 😊
@atruskie if you have time it would be great to get your eyes over the change to `TryFindMappingEntry` and how I have described it's functionality in the xml doc. I've given it a go with my understanding but I may have missed a quirk of it's original design/intention.

YamlDotNet/Core/ParserExtensions.cs
```diff
@@ -147,5 +147,63 @@ public static void SkipThisAndNestedEvents(this IParser parser)
...
+            if (parser.TryConsume<MappingStart>(out var _start))
+            {
+                while (parser.Current != null)
+                {
+                    // so we only want to check keys in this mapping, don't descend
+                    switch (parser.Current)
+                    {
+                        case Scalar scalar:
+                            // we've found a scalar, check if it's value matches one
+                            // of our predicate
```
atruskie, line 173:
```suggestion
                            // we've found a scalar, check if its value matches our predicate
```

YamlDotNet/Core/ParserExtensions.cs
```diff
@@ -147,5 +147,63 @@ public static void SkipThisAndNestedEvents(this IParser parser)
         {
             return Accept<T>(parser, out var _);
         }
+
+        /// <summary>
+        /// Checks whether the current event contains a mapping entry that returns true for the specified selector.
+        /// If one is found, it's key is returned as a Scalar and it's value as a ParsingEvent, the mapping entry
+        /// is consumed and the parser is moved to the next event, and the whole function returns true.
+        /// If the event is not a mapping event or a matching key is not found, returns false.
+        /// </summary>
```
atruskie, line 156:
```suggestion
         /// Attempts to find a key on a YAML mapping that matches our predicate.
         /// This is useful for scanning a mapping for type discriminator information.
         /// For example: looking for a `kind` key on an object.
         ///
         /// This function only checks mappings, and only looks at the current depth.
         ///
         /// If the event is a mapping and has a key that satisfies the predicate the scan 
         /// will stop, return true, and set
         /// <paramref>key</paramref> and <paramref>value</paramref>.
         ///
        /// If the event is not a mapping event or a matching key is not found, returns false.
```

Your description is correct but perhaps a little jargony. I've added my suggestion, feel free to adapt how you like.

YamlDotNet/Serialization/BufferedDeserialization/BufferedNodeDeserializer.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// If any of it's <see cref="ITypeDiscriminator" /> has a matching BaseType, the BufferedNodeDeserializer will
+    /// begin buffering the yaml stream. It will then use the matching <see cref="ITypeDiscriminator" />s to determine
+    /// a dotnet output type for the yaml node. As the node is buffered, the <see cref="ITypeDiscriminator" />s are
+    /// able to examine the actual values within, and use these when discriminating a type.
+    /// Once a matching type is found, the BufferedNodeDeserializer uses it's inner deserializers to perform
+    /// the final deserialization for that type & object.
+    /// Usually you will want all default <see cref="INodeDeserializer" />s that exist in the outer
+    /// <see cref="Deserializer" /> to also be used as inner deserializers.
+    /// </summary>
+    public class BufferedNodeDeserializer : INodeDeserializer
```
atruskie, line 23: Perhaps this type name should be changed to `TypeDiscriminatingNodeDeserializer`. 

It reflects its actual purpose (discriminating types) rather than describing its method of action (buffering).

atruskie (commented):
This is fantastic!

A few notes:

- Your tests don't cover a fall through case (i.e. two registrations of `AddKeyValueTypeDiscriminator` or `AddUniqueKeyTypeDiscriminator`) but the code seems to support it. I.e. would this work:

```
            var bufferedDeserializer = new DeserializerBuilder()
                .WithBufferedNodeDeserializer(options => {
                    options.AddKeyValueTypeDiscriminator<KubernetesResource>(
                        "kind",
                        new Dictionary<string, Type>()
                        {
                            { "Namespace", typeof(KubernetesNamespace) },
                        })
                       .AddKeyValueTypeDiscriminator<KubernetesResource>(
                        "kind",
                        new Dictionary<string, Type>()
                        {
                            { "Service", typeof(KubernetesService) }
                        });
                    },
                    maxDepth: 3,
                    maxLength: 40)
                .Build();
```

- The name `BufferedNodeDeserializer` is probably one I would change (I know I chose the name originally, my bad) since it doesn't describe what it's doing but rather how - and how is unimportant. Suggestion: `TypeDiscriminatingNodeDeserializer`.

I'm so hyped you did the work for this. There's no way I would have had the capacity, so thank you!

JJ11teen:
Thanks @atruskie! Have made those changes (:

JJ11teen:
Rebased onto master again (:

YamlDotNet/Serialization/DeserializerBuilder.cs
```diff
@@ -246,6 +247,28 @@ public DeserializerBuilder WithoutNodeDeserializer(Type nodeDeserializerType)
...
+        /// <param name="maxDepth">Configures the max depth of yaml nodes that will be buffered. A value of -1 (the default) means yaml nodes of any depth will be buffered.</param>
+        /// <param name="maxLength">Configures the max number of yaml nodes that will be buffered. A value of -1 (the default) means there is no limit on the number of yaml nodes buffered.</param>
+        public DeserializerBuilder WithTypeDiscriminatingNodeDeserializer(
+            Action<ITypeDiscriminatingNodeDeserializerOptions> configureTypeDiscriminatingNodeDeserializerOptions, int maxDepth = -1, int maxLength = -1)
+        {
+            var options = new TypeDiscriminatingNodeDeserializerOptions();
+            configureTypeDiscriminatingNodeDeserializerOptions(options);
+            // We use all current NodeDeserializers as the inner deserializers for the TypeDiscriminatingNodeDeserializer,
+            // so that it can successfully deserialize anything our root deserializer can.
+            var TypeDiscriminatingNodeDeserializer = new TypeDiscriminatingNodeDeserializer(nodeDeserializerFactories.BuildComponentList(), options.discriminators, maxDepth, maxLength);
```
EdwardCooke, line 265: This should be a lower case `T` on the variable name.

YamlDotNet/Serialization/StaticDeserializerBuilder.cs
```diff
@@ -205,6 +206,28 @@ public StaticDeserializerBuilder WithoutNodeDeserializer(Type nodeDeserializerTy
...
+        /// <param name="maxDepth">Configures the max depth of yaml nodes that will be buffered. A value of -1 (the default) means yaml nodes of any depth will be buffered.</param>
+        /// <param name="maxLength">Configures the max number of yaml nodes that will be buffered. A value of -1 (the default) means there is no limit on the number of yaml nodes buffered.</param>
+        public StaticDeserializerBuilder WithTypeDiscriminatingNodeDeserializer(
+            Action<ITypeDiscriminatingNodeDeserializerOptions> configureTypeDiscriminatingNodeDeserializerOptions, int maxDepth = -1, int maxLength = -1)
+        {
+            var options = new TypeDiscriminatingNodeDeserializerOptions();
+            configureTypeDiscriminatingNodeDeserializerOptions(options);
+            // We use all current NodeDeserializers as the inner deserializers for the TypeDiscriminatingNodeDeserializer,
+            // so that it can successfully deserialize anything our root deserializer can.
+            var TypeDiscriminatingNodeDeserializer = new TypeDiscriminatingNodeDeserializer(nodeDeserializerFactories.BuildComponentList(), options.discriminators, maxDepth, maxLength);
```
EdwardCooke, line 224: Lowercase `T` here as well.

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminatingNodeDeserializer.cs
```diff
@@ -0,0 +1,99 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/UniqueKeyTypeDiscriminator.cs
```diff
@@ -0,0 +1,67 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet.Test/Serialization/BufferedDeserialization/KeyValueTypeDiscriminatorTests.cs
```diff
@@ -0,0 +1,244 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet.Test/Serialization/BufferedDeserialization/TypeDiscriminatingNodeDeserializerTests.cs
```diff
@@ -0,0 +1,68 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet.Test/Serialization/BufferedDeserialization/UniqueKeyTypeDiscriminatorTests.cs
```diff
@@ -0,0 +1,82 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet/Serialization/BufferedDeserialization/ITypeDiscriminatingNodeDeserializerOptions.cs
```diff
@@ -0,0 +1,26 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet/Serialization/BufferedDeserialization/ParserBuffer.cs
```diff
@@ -0,0 +1,71 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminatingNodeDeserializerOptions.cs
```diff
@@ -0,0 +1,46 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/IValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,30 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminators/KeyValueTypeDiscriminator.cs
```diff
@@ -0,0 +1,75 @@
+using System;
```
EdwardCooke, line 1: Missing required file header

YamlDotNet/Serialization/DeserializerBuilder.cs
```diff
@@ -25,6 +25,7 @@
 using System.Diagnostics.CodeAnalysis;
```
EdwardCooke, line 25: Missing required file header

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminatingNodeDeserializer.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace YamlDotNet.Serialization.BufferedDeserialization
+{
+    /// <summary>
+    /// The TypeDiscriminatingNodeDeserializer acts as a psuedo <see cref="INodeDeserializer" />.
+    /// If any of it's <see cref="ITypeDiscriminator" /> has a matching BaseType, the TypeDiscriminatingNodeDeserializer will
+    /// begin buffering the yaml stream. It will then use the matching <see cref="ITypeDiscriminator" />s to determine
+    /// a dotnet output type for the yaml node. As the node is buffered, the <see cref="ITypeDiscriminator" />s are
+    /// able to examine the actual values within, and use these when discriminating a type.
+    /// Once a matching type is found, the TypeDiscriminatingNodeDeserializer uses it's inner deserializers to perform
+    /// the final deserialization for that type & object.
```
EdwardCooke, line 19: This should be `&amp;` and not just an `&` sign

YamlDotNet/Serialization/BufferedDeserialization/TypeDiscriminatingNodeDeserializer.cs
```diff
@@ -0,0 +1,99 @@
...
+                if (deserializer.Deserialize(buffer, actualType, nestedObjectDeserializer, out value))
+                {
+                    return true;
+                }
+            }
+
+            value = null;
+            return false;
+        }
+    }
```
EdwardCooke, line 98: I know its a little nit picky, but it throws a warning on build because of the extra whitespace after the curly brace here.

YamlDotNet/Core/ParserExtensions.cs
```diff
@@ -147,5 +147,70 @@ public static void SkipThisAndNestedEvents(this IParser parser)
...
+        /// be consumed.
+        ///
+        /// If the event is not a mapping event or a matching key is not found, returns false.
+        /// </summary>
+        /// <param name="selector">The selector to filter the mapping by.</param>
+        /// <param name="key">The matching key of the mapping as a Scalar, or null if no matching key found</param>
+        /// <param name="value">The matching value of the mapping as a ParsingEvent, or null if no matching key found</param>
+        /// <returns>Returns true if the current event is a mapping entry with a key that matches the selector;
+        /// otherwise returns false.</returns>
+        public static bool TryFindMappingEntry(this IParser parser, Func<Scalar, bool> selector, [MaybeNullWhen(false)] out Scalar? key, [MaybeNullWhen(false)] out ParsingEvent? value)
```
EdwardCooke, line 170: The `parser` parameter here needs an xml comment above.

EdwardCooke:
Once the latest comments are fixed we should be good to merge this in.

YamlDotNet/Serialization/DeserializerBuilder.cs
JJ11teen: This one already exists 😅 Thanks for going through and being explicit with what you want though, I appreciate it 😊

JJ11teen:
All done @EdwardCooke! (:

EdwardCooke:
I’ll merge this in when I get time, should be this week.

pmikstacki:
Hi! When it will be available as nuget package?

aaubry:
This feature has been released in [version 13.1.0](https://github.com/aaubry/YamlDotNet/releases/tag/v13.1.0).

rcdailey:
I have just pulled down the 13.1.0 release and tried using this, but the lack of documentation is killing me.

- This PR didn't serve as a helpful form of documentation due to the change in class names over time
- There's no documentation markdown anywhere that has tutorials or the like to help people get started with this.
- Some public interfaces, like `ITypeDiscriminatingNodeDeserializerOptions`, nave no C# documentation so it's unclear, for example, what the difference is between `AddUniqueKeyTypeDiscriminator` and `AddKeyValueTypeDiscriminator`.

Don't get me wrong, I'm not here just to complain. Just pointing out something that I think will make it difficult for users to adopt this new feature. I'm sure if I swim into the unit tests I'll eventually figure it out, but hours are pretty valuable these days...

In any case, I appreciate the months you folks spent on getting this ready.

EdwardCooke:
Thanks for the feedback, I'll work on building out some documentation in the wiki, I know this particular feature was a bit complex in nature. I'll also work on getting those missing XML comments on the interface.

EdwardCooke:
Here's a little wiki page documenting how to use this new feature

https://github.com/aaubry/YamlDotNet/wiki/Deserialization---Type-Discrimators

R3ven6eR (commented):
- [ ]

rcdailey:
> Here's a little wiki page documenting how to use this new feature
> 
> [aaubry/YamlDotNet/wiki/Deserialization---Type-Discrimators](https://github.com/aaubry/YamlDotNet/wiki/Deserialization---Type-Discrimators)

For future visitors, this is the correct link: https://github.com/aaubry/YamlDotNet/wiki/Deserialization---Type-Discriminators

Thanks again for the examples!

PR #778, opened 2023-02-13
Quoted scalars are flow scalars only if flowLevel > 0

rutger-dijkstra:
The parser currently crashes on (some) YAML generated by Ruby.

Serializing objects to YAML in Ruby generally results in keys starting with a colon (symbols) and in single quoted values if strings values could otherwise be mistaken for numbers. Something like:
```
:string_value: '1'
:numeric_value: 2
```
The scanner will, after fetching the quoted scalar `'1'`, gobble up the initial colon from `:numeric_value` and insert another block level. The parser then raises the exception `YamlDotNet.Core.SemanticErrorException : While parsing a block mapping, did not find expected key.`

This fix remedies that behavior by not setting `flowScalarFetched` to `true` after fetching a quoted scalar on the block level

### DIFF
--- a/YamlDotNet.Test/Core/ScannerTests.cs
+++ b/YamlDotNet.Test/Core/ScannerTests.cs
@@ -443,6 +443,26 @@ namespace YamlDotNet.Test.Core
                 Error("Invalid key indicator format."));
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


aaubry:

PR #786, opened 2023-03-09
Fix NRT for ISerializer

meziantou:
Serializing `null` is allowed, but the code is not correctly annotated.


EdwardCooke:
Can you add a test to make sure that passing null to this method works as expected and continues to work as expected in the future?

meziantou:
I've added a test and enable nullable annotation in this test to also validate the nullable annotation.

EdwardCooke:
Thanks. I'll take a closer look at this tonight and most likely merge it in. First glance it looks good.

aaubry:
This feature has been released in [version 13.0.2](https://github.com/aaubry/YamlDotNet/releases/tag/v13.0.2).

Issue #803, opened 2023-04-27
How to serialize properties with blank lines between them


emanuel-v-r:
How to place some blank lines between properties.
In my case this is helpful for separating sections of the yaml.

Example:

Considering this class:
```
public class MyYaml
{
    public Section Section1 { get; set; }

    public Section Section2 { get; set; }
}
```

I am getting this:
```
section1: 
  name: section 1 name
section2: 
  name: section 2 name
```

But I want this:
```
section1: 
  name: section 1 name

section2: 
  name: section 2 name
```
Tried to emmit a Scalar event with a break line, but it does not work properly as introduces some other chars.
I guess that there might be a cleaner way of doing this

EdwardCooke:
Checkout my answer to #871 I'm going to close this in favor of that one I just linked.

Issue #523, opened 2020-08-24
While scanning a literal block scalar, found extra spaces in first line.


LaughingJohn:
Hi,

I have some YAML files from a 3rd party which I'm reading and converting to JSON to make it easier to process. For a few files on the deserialize step it is failing with "While scanning a literal block scalar, found extra spaces in first line".

Trace:

```
at YamlDotNet.Core.Scanner.ScanBlockScalarBreaks(Int32 currentIndent, StringBuilder breaks, Boolean isLiteral, Mark& end, Nullable`1& isFirstLine)   at YamlDotNet.Core.Scanner.ScanBlockScalar(Boolean isLiteral)   at YamlDotNet.Core.Scanner.FetchBlockScalar(Boolean isLiteral)
```

Code:
```
var yamlDeserializer = new DeserializerBuilder().Build();
var yamlObject = yamlDeserializer.Deserialize(sr); -- FAILS HERE --
```

Looking at the data it appears that the problem is the files have a multi-line literal with additional carriage returns at the beginning. I'm new to YAML, but I'm wondering why it isn't considered reasonable to have more than one CRLF or indeed more than one space at the start of some literal text? Apart from editing the files is there any way around this?

Example literal with extra CRLF at the start:

```
    Body: |+
      
      Begin forwarded message:
```

I presume it is to do with trying to establish the indentation - are the files invalid or should the scanner be reading until it finds a non-blank line?

Thanks.

wspresto:
I think this issue warrants a response. We are grappling with the same problem manifesting in a different way. Whenever we have empty lines (auto indented by text editors) our literal style is auto-forced into double quoted. I found where this is happening in the code Ln 911 @ Emitter.cs .

Rather than throwing an error, the Emitter object quietly changes the style to double quote if it feels the Scalar in question does not fit the bill to be a block Literal type. .. Suggestion: Allow forced styles or throw an error if requested style is not allowed.

aaubry:
Of course a response is warranted. This is certainly a bug but I didn't have a chance to look into it. Do you want to help, @wspresto  ?

LaughingJohn:
[...]

PR #804, opened 2023-04-29
Allow to Serialize with line breaks between sections

emanuel-v-r:
This is related to the issue that I described here https://github.com/aaubry/YamlDotNet/issues/803.
For my usage it's important to be able to serialize a Yaml with line breaks.
Due to yaml's nature, which is declarative and human readable I think it makes sense to support it. We often see files like docker-compose  separating sections by break line.
With this PR, that would be possible, either by using YamlMemberAttribute, or by emiting the event manually.
Please let me know if there's an existing and more straightforward way of doing this, then I will abort my PR.
Thanks.

### DIFF
--- a/YamlDotNet.Test/Serialization/YamlNewLineTests.cs
+++ b/YamlDotNet.Test/Serialization/YamlNewLineTests.cs
@@ -1,0 +1,124 @@
+﻿// This file is part of YamlDotNet - A .NET library for YAML.
+// Copyright (c) Antoine Aubry and contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+// of the Software, and to permit persons to whom the Software is furnished to do
+// so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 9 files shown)


emanuel-v-r:
@EdwardCooke, would it be possible to get someone to review this?
It would be really helpful.
Thank you.

JuergenGutsch:
Maybe also related to #523? I will have a look

emanuel-v-r:
> Maybe also related to #523? I will have a look

Actually this is more related to serialization, will update the title and description

EdwardCooke:
I’ll try and take a look this weekend.

EdwardCooke:
Sorry for the long delay, was on vacation :) Anyways, what I would prefer is a way of adding any arbitrary content before it, that way it can be whatever is desired by the developer, if we did this as a string, then people could put in whatever the feel like, including something like

```yaml
something:

# i like
# new lines!

  like: that

  or: just a new line
```

Also, you'll need to do this on a branch in your fork named something other than `master`. Using `master` in your fork breaks the build.

emanuel-v-r:
> 

I have mixed feelings about that, I mean it would be more flexible for sure, but at the same time it would allow you to serialize an invalid yaml, while this way it would be.more restricted.
Nevertheless if you prefer it that way, I have no issues, it will fit our scenario as well.
Let me know, and I will change the branch and adapt it accordingly

EdwardCooke:
I’ve been thinking about this more. I really think having the ability to put whatever a person wants in there is a good thing due to flexibility. One idea I have is maybe a check in the attribute setter that makes sure every line is either empty or starts with spaces and/or pound sign. That would guarantee a valid yaml structure at compile time, in theory. I’d like that tested to make sure. But at least it would be validated at runtime at the very least.

EdwardCooke:
Another issue here is that your branch shouldn't be on master, please create your own, otherwise the build will fail.

EdwardCooke:
Are you going to create your own, non-master, branch? Or can I close this?

emanuel-v-r:
I am closing this PR as the approach that you want to follow is a more flexible one and I am not actually sure that those changes are as simple to implement, in the way that I am afraid of users breaking the seriaization/deserialization due to such flexibility.
Probably that would require more extensive testing as well.

PR #808, opened 2023-05-28
Fix infinite recursion of the hash code function of YamlMappingNode

tomer-amir:
I have an issue where the hash code function enters infinite recursion in some cases with circular references.
This is a suggestion to fix the issue, as I noticed that the start position seems unique, so that can be relied on, and this does not require any recursion.

Please tell me if you think that there is a better fix, but this worked for me when I tested it.

Here is an example that causes infinite recursion:

```yaml
? !Key
    properties:
            properties:
                rowID: &id1 !Key
                        rowID: *id1
: *id2
```

I am not sure what exactly causes the issue here, this is a filtered snippet from a YAML I was trying to parse and I got the issue with, so it might not be exactly valid, but it reproduces the issue and verifies the fix.


EdwardCooke:
We need some tests for the fix to make sure we don't accidentally break it in the future. If you can add those we can look at merging it in.

barapiiro:
> We need some tests for the fix to make sure we don't accidentally break it in the future. If you can add those we can look at merging it in.

Hi, I added some test cases that work only after the fix. 
we would appreciate it if you'll take another look and merge it 🙏🏽

aaubry:
This feature has been released in [version 15.1.6](https://github.com/aaubry/YamlDotNet/releases/tag/v15.1.6).

Issue #812, opened 2023-05-31
Emitter doesn't handle comments between the mapping value and key


YihaoOct:
**Describe the bug**

When there is a comment after a mapping key and before the value, the Emitter doesn't generate the correct YAML output.

Examples are:

Inline comment:

```yaml
hello: # comment
  - world
```

Parse it and emit it again, the output will be:
```
hello # this is a comment:
- world
```

Standalone comment:

```yaml
hello:
  # comment
  - world
```

Parse it and emit it again, the output will be:
```
hello
# this is a comment
:
- world
```

**To Reproduce**

Checkout the branch for this PR: https://github.com/aaubry/YamlDotNet/pull/811

Run and debug the two new test cases, see the broken output.

EdwardCooke:
Well that would be an annoying bug....I'll take a look at this and see about merging it in today.

PR #814, opened 2023-06-03
Fix Emitter to handle comments between a mapping key and a mapping/sequence value

YihaoOct:
This PR fixes https://github.com/aaubry/YamlDotNet/issues/812

TL;DR the Emitter doesn't handle comment events that are emitted between a mapping key and a mapping value of the mapping or sequence type. So parsing and regenerate the following yaml will fail for example:

```yaml
hello:
  # comment
  - world
```

becomes
```yaml
hello
# this is a comment
:
- world
```

This PR moves the writing of the `:` symbol to be inside the block mapping key emission. Previously it was at the beginning of the block mapping value emission. If the mapping key is prefixed with a `?` (explicit form), however, the `:` is still kept at the mapping value to preserve the existing behaviour, since in the explicit form, the comment does not break YAML syntax. That is,
```yaml
? hello
# this is a comment
:
- world
```
is valid.

### DIFF
--- a/YamlDotNet.Test/Core/EmitterTests.cs
+++ b/YamlDotNet.Test/Core/EmitterTests.cs
@@ -297,6 +297,28 @@ namespace YamlDotNet.Test.Core
         }
 
         [Fact]
+        public void CommentsBetweenMappingKeyAndValueAreEmittedCorrectly()
+        {
+            var events = MappingWith(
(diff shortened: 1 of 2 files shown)


aaubry:
This feature has been released in [version 13.1.1](https://github.com/aaubry/YamlDotNet/releases/tag/v13.1.1).

Issue #820, opened 2023-06-19
Cannot parse YAML file, (Line: 2, Col: 3, Idx: 6) - (Line: 2, Col: 3, Idx: 6): Exception during deserialization


rudolfolah:
**Describe the bug**

I'm encountering an issue while trying to load an Ansible task file, where I receive an exception related to deserialization.

When attempting to load the Ansible task file, the following exception arises:

```
Unhandled exception. (Line: 2, Col: 3, Idx: 6) - (Line: 2, Col: 3, Idx: 6): Exception during deserialization
```

**To Reproduce**

Here is the specific C# code I am executing:

```csharp
string tasksMainYaml = File.ReadAllText("../../../sample_data/tasks_main.yml");
var yamlDeserializer = new DeserializerBuilder().Build();
var parser = new Parser(new StringReader(tasksMainYaml));
YamlDotNet.RepresentationModel.YamlStream yaml = new YamlDotNet.RepresentationModel.YamlStream();
yaml.Load(new StringReader(tasksMainYaml));
parser.Expect<StreamStart>();

while (parser.Accept<DocumentStart>())
{
    // Deserialize the document
    var doc = yamlDeserializer.Deserialize<List<string>>(parser);

    Console.WriteLine("## Document");
    foreach (var item in doc)
    {
        Console.WriteLine(item);
    }
}
```

The YAML file that I am trying to parse includes the following:

```yml
---
- name: Install certbot
  tags: setup
  community.general.snap:
    name: certbot
    classic: true
- name: Create symlink for Certbot binary
  tags: setup
  ansible.builtin.file:
    src: /snap/bin/certbot
    dest: /usr/bin/certbot
    state: link
  become: true
- name: Check if the certificate was already generated
  tags: generate_cert
  ansible.builtin.stat:
    path: "/etc/letsencrypt/live/{{ domain }}/fullchain.pem"
  register: cert_fullchain_file
- name: Generate certificate manually
  tags: generate_cert
  ansible.builtin.command:
    cmd: "certbot certonly --standalone --preferred-challenges http --agree-tos --email {{ email }} -d {{ domain }}"
    creates: "/etc/letsencrypt/live/{{ domain }}/fullchain.pem"
  when: not cert_fullchain_file.stat.exists
```

Any assistance in troubleshooting this issue would be greatly appreciated.

EdwardCooke:
You're trying to deserialize a mapping into a string which isn't going to work. If instead you did this in your while loop:
[...]

PR #821, opened 2023-06-21
Deserialize generic objects error message and sample

rudolfolah:
Here is one update based on https://github.com/aaubry/YamlDotNet/issues/820

* added an example to show how to deserialize data into generic objects like lists, dictionaries, or strings, etc.

### DIFF
--- a/YamlDotNet.Samples/DeserializeGenericObject.cs
+++ b/YamlDotNet.Samples/DeserializeGenericObject.cs
@@ -1,0 +1,89 @@
+﻿// This file is part of YamlDotNet - A .NET library for YAML.
+// Copyright (c) Antoine Aubry and contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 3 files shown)


EdwardCooke:
One issue I have with changing that exception message is that it can be caused by reasons other than the data type, adding that in there could be just as misleading as it not being there. A simple example:
```csharp
var yaml = @"
Test: x";
var test = deserializer.Deserialize<Foo>(yaml);
Console.WriteLine(test.Test);

class Foo
{
    private string? _test;

    public string Test
    {
        get
        {
            return _test ?? string.Empty;
        }
        set
        {
            throw new Exception("Setter is broken");
        }
    }
}
```

It's not the data type that's broken, but instead the property itself. Yes, that exception would be in the inner exception property, but it becomes confusing with the top exception message saying one thing but it being completely different.

rudolfolah:
That's a fair point, I can revert that part of the PR. Let me know if there's anything else needed for the sample code and I'll be happy to update the PR further.

EdwardCooke:
A I think reverting the exception messages is all.

rudolfolah:
Awesome, reverted it and updated the PR description.

aaubry:

PR #825, opened 2023-07-06
Use a span to reduce string allocations

rbnswartz:
On versions that support it we can use Span<char> to ease GC pressure by not having to allocate a temporary string from substring, instead we can just get a readonly span of what we want and send that to the string builder

### DIFF
--- a/YamlDotNet/Core/Scanner.cs
+++ b/YamlDotNet/Core/Scanner.cs
@@ -2405,7 +2405,11 @@ namespace YamlDotNet.Core
 
             if (head != null && head.Length > 1)
             {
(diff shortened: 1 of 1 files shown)


rbnswartz:
There wasn't anything in the style guide about # directives so if you would like me to adjust the indentation on them let me know.

EdwardCooke:
Thanks for this. I’ll review shortly, I’m thinking it may be beneficial to somehow break this out so it’s reusable to not litter the code with compiler directives. We do a lot like this throughout the library and it would probably help.

aaubry:

Issue #315, opened 2018-04-10
surround value with double quotes


dragouf:
I have number which I need to keep as string in my yaml and so I need to surround them with quotes.

I tried to find how to achieve that with your libraries but I didn't found how to achieve that.

Is there a way to do that ?

aaubry:
You can use the `YamlMember` attribute to control this:

```c#
public class QuotedInt
{
	[YamlMember(ScalarStyle = ScalarStyle.DoubleQuoted)]
	public int Value { get; set; }
}
```

[example](https://dotnetfiddle.net/LrfVZ4)

dragouf:
Thanks, unfortunately I serialize a dynamic object and I can't do this.

I ended up doing something like this but I'm really not sure it's the best solution especially the typeof(Object) condition :

```
public class QuoteSurroundingEventEmitter : ChainedEventEmitter
{
    public QuoteSurroundingEventEmitter(IEventEmitter nextEmitter)  : base(nextEmitter)
    { }

    public override void Emit(ScalarEventInfo eventInfo, IEmitter emitter)
    {
        if(eventInfo.Source.StaticType == typeof (Object))
            eventInfo.Style = ScalarStyle.DoubleQuoted;
            base.Emit(eventInfo, emitter);
    }
}
```

and use it like this :
```
var serializer = new SerializerBuilder()
                .WithEventEmitter(nextEmitter => new QuoteSurroundingEventEmitter(nextEmitter))
                .Build();
```

lvpeng2016:
cool

daniol:
Use WithQuotingNecessaryStrings():
`var serializer = new YamlDotNet.Serialization.SerializerBuilder().WithQuotingNecessaryStrings().Build();`

PR #827, opened 2023-07-17
Add default scalar style

### LINKED ISSUE #714: Wrap all string values in quotes when serializing, if possible
I'd like to wrap string values in quotes during serialization, if it's possible to do so safely. YamlDotNet is being used to handle configuration files, and one of the biggest issues I've encountered is that the default value for a key is serialized without quotes, so if a user modifies it to include a character which requires surrounding quotes, it results in a parsing error. I've looked into #315, although it seems as though the method described there doesn't give as much control as is desired and surrounds keys in quotes as well as values. Any help with solving this would be greatly appreciated!

For reference, this is for https://github.com/Exiled-Team/EXILED/issues/563.

tymokvo:
## Summary
Re: #714 , it is sometimes desirable to force a quotation style on all strings in a YAML document. For example, when some strings contain valid numeric literals.

This implements the suggested change from that issue by exposing the "default scalar style" as a constructor argument. Previously, this value was hard-coded to `ScalarStyle.Any` as the local value `suggestedStyle`.

## Changes
- `YamlDotNet/Serialization`
  - Add `defaultScalarStyle = ScalarStyle.Any` to `TypeAssigningEventEmitter`
  - Add `SerializerBuilder.WithDefaultScalarStyle`
    - I wasn't sure of the best way to create the `TypeAssigningEventEmitter`, so I copied the instantiation from the constructor and used `loc.InsteadOf<_>` to replace it.
    - I'm not sure how necessary it is, but here it would have been a bit more ergonomic to have an instance member on the `SerializerBuilder` to construct its `TypeAssigningEventEmitter`
- `YamlDotNet.Test`
  - Add `SerializationRespectsDefaultScalarStyle`
    - This is only testing `SingleQuote`. I can add others if necessary.
  - Add  `MixedFormatScalarStyleExample` to helpers

## To Do
- [x] Add `StaticSerializerBuilder.WithDefaultScalarStyle`
- [x] Use literal strings in test case
- [x] Increase newline consistency quotient
- [x] Add new constructors instead of changing signatures

Fix #714

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -456,6 +456,19 @@ namespace YamlDotNet.Test.Serialization
         public string DoubleQuotedString { get; set; }
     }
 
+    /// <summary>
+    /// This class demonstrates an object with serializable data that will be
+    /// written to YAML with strings that are formatted as valid numbers.
+    /// </summary>
+    public class MixedFormatScalarStyleExample
+    {
+        public string[] Data { get; }
+        public MixedFormatScalarStyleExample(string[] data)
+        {
+            Data = data;
+        }
+    }
+
     public class DefaultsExample
     {
         public const string DefaultValue = "myDefault";
--- a/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTestHelper.cs
@@ -463,6 +463,7 @@ namespace YamlDotNet.Test.Serialization
     public class MixedFormatScalarStyleExample
     {
         public string[] Data { get; }
+
         public MixedFormatScalarStyleExample(string[] data)
         {
             Data = data;
--- a/YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs
+++ b/YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs
@@ -65,14 +65,14 @@ namespace YamlDotNet.Serialization.EventEmitters
                 + @"|\.(nan|NaN|NAN)"
             + @")$";
 
-        private readonly ScalarStyle? globalScalarStyle;
+        private readonly ScalarStyle defaultScalarStyle;
 
-        public TypeAssigningEventEmitter(IEventEmitter nextEmitter, bool requireTagWhenStaticAndActualTypesAreDifferent, IDictionary<Type, TagName> tagMappings, bool quoteNecessaryStrings, bool quoteYaml1_1Strings, ScalarStyle? globalScalarStyle = null)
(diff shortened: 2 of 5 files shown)


tymokvo:
Hm, I don't know enough about CI here to fix the failure. Seems like an upload failed?
```console
Error uploading artifact to the storage: Remote server returned 500: Internal Server Error
```

YamlDotNet.Test/Serialization/SerializationTestHelper.cs
```diff
@@ -456,6 +456,19 @@ public ScalarStyleExample()
...
     }
 
+    /// <summary>
+    /// This class demonstrates an object with serializable data that will be
+    /// written to YAML with strings that are formatted as valid numbers.
+    /// </summary>
+    public class MixedFormatScalarStyleExample
+    {
+        public string[] Data { get; }
+        public MixedFormatScalarStyleExample(string[] data)
```
EdwardCooke, line 466: I know it's nit picky, especially in a test class, but can you add an empty line between the property and the method? That's the general standard throughout this library.

YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs
```diff
@@ -65,6 +65,20 @@ public sealed class TypeAssigningEventEmitter : ChainedEventEmitter
...
+        {
+            this.defaultScalarStyle = defaultScalarStyle;
+        }
+
+        public TypeAssigningEventEmitter(IEventEmitter nextEmitter, bool requireTagWhenStaticAndActualTypesAreDifferent, IDictionary<Type, TagName> tagMappings, bool quoteNecessaryStrings, ScalarStyle defaultScalarStyle)
+            : this(nextEmitter, requireTagWhenStaticAndActualTypesAreDifferent, tagMappings, quoteNecessaryStrings)
+        {
+            this.defaultScalarStyle = defaultScalarStyle;
+        }
+
```
tymokvo, line 81: I ended up adding 2 constructors since `SerializerBuilder` and `StaticSerializerBuilder` rely on 2 different existing constructors. Is that the right approach here?

YamlDotNet.Test/Serialization/SerializationTestHelper.cs
tymokvo: Added!

YamlDotNet/Serialization/EventEmitters/TypeAssigningEventEmitter.cs
EdwardCooke: Good enough :)

EdwardCooke (approved):
Looks good.

aaubry:

daniol:
This quotes everything, also keys, not only values. Hence this is not what was in #714 or #315 requested.

PR #834, opened 2023-08-22
Support null values in YamlScalarNode for YamlStream

### LINKED ISSUE #810: `YamlStream.Load(TextReader)` interprets an empty value as an empty string instead of null.
**Describe the bug**

Version: 13.1.0

According to:
- https://yaml.org/type/null.html
- https://yaml.org/spec/1.2.2/#72-empty-nodes

```yaml
annotations:
```

generally expected to be equivalent to 

```yaml
annotations: ~
```

However, `YamlStream.Load(TextReader)` defaults `annotations` to the empty string. i.e.,

```yaml
annotations: ''
```

**To Reproduce**

```cs
using System;
using System.IO;
using YamlDotNet.RepresentationModel;
					
public class Program
{
	public static void Main()
	{
		var inYaml = "annotations:";

                using var sr = new StringReader(inYaml);

                yamlStream.Load(sr);

                using var writer = new StringWriter();
                new YamlStream(yamlStream.Documents).Save(writer);
                var outYaml = writer.ToString();
		
		Console.Write(outYaml);
	}
}
```

**Actual**
```
annotations: ''
...

```

**Expected**
```
annotations: ~
...

```
or
```
annotations: null
...

```

EdwardCooke:
Fixes #810

For backwards compatibility we wont set the `Value` property to `null`.

### DIFF
--- a/YamlDotNet.Test/RepresentationModel/YamlStreamTests.cs
+++ b/YamlDotNet.Test/RepresentationModel/YamlStreamTests.cs
@@ -27,7 +27,9 @@ using System.Text;
 using FluentAssertions;
 using Xunit;
 using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
 using YamlDotNet.RepresentationModel;
+using YamlDotNet.Serialization.Schemas;
 
 namespace YamlDotNet.Test.RepresentationModel
 {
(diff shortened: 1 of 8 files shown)


aaubry:
This feature has been released in [version 13.3.1](https://github.com/aaubry/YamlDotNet/releases/tag/v13.3.1).

PR #835, opened 2023-08-22
Set IsKey on quoted scalar keys

### LINKED ISSUE #817: Parser does not recognize quoted keys
I'm using the Parser class to read in YAML content.
When the Parser instance encounters a mapping key which is double quoted or single quoted, it returns a Scalar ParsingEvent in the Current property (it is OK), but fails to set IsKey true (this is the bug).

So parsing up YAML content, like:
```
"name": "John"
```
fails as "name" is recognized as plain scalar and not a key. This makes parsing JSON content impossible for me, as correct JSON like
```
{ "name": "John" }
```
also fails.

If I remove the single or double quotes from around the mapping keys, everything works just fine.

The issue is present in releases 13.1.0 and 13.0.2 at least.

EdwardCooke:
Fixes #817

### DIFF
--- a/YamlDotNet/Core/Scanner.cs
+++ b/YamlDotNet/Core/Scanner.cs
@@ -1785,9 +1785,10 @@ namespace YamlDotNet.Core
             flowScalarFetched = flowLevel > 0;
 
             // Create the SCALAR token and append it to the queue.
+            var scalar = ScanFlowScalar(isSingleQuoted);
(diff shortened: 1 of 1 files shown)


aaubry:
This feature has been released in [version 13.3.1](https://github.com/aaubry/YamlDotNet/releases/tag/v13.3.1).

Issue #851, opened 2023-10-05
Add support for including types defined in another assembly in static code generation


prochnowc:
**Is your feature request related to a problem? Please describe.**
My serializable types are defined in some 3rd party assembly and I can not add the `[YamlSerializable]` attribute.

**Describe the solution you'd like**
I would like the static code generator to include types defined in another assembly.

EdwardCooke:
Since this is now complete, I'm closing the issue. Thank you for taking care of it so quickly.

PR #852, opened 2023-10-05
Add support for including types defined in another assembly in static code generation

prochnowc:
This PR adds a second constructor to the `YamlSerializableAttribute` to allow static code generation for types defined in another assembly.

Fixes issue #851

### DIFF
--- a/YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
+++ b/YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
@@ -66,17 +66,17 @@ namespace YamlDotNet.Analyzers.StaticGenerator
             }
         }
 
-        private void AddSerializableClass(INamedTypeSymbol classSymbol)
+        private void AddSerializableClass(INamedTypeSymbol? classSymbol)
         {
             ClassObject classObject;
-            var className = SanitizeName(classSymbol.GetFullName());
+            var className = SanitizeName(classSymbol!.GetFullName());
             if (Classes.ContainsKey(className))
             {
                 classObject = Classes[className];
             }
             else
             {
-                classObject = new ClassObject(className, classSymbol);
+                classObject = new ClassObject(className, classSymbol!);
                 Classes[className] = classObject;
             }
             while (classSymbol != null)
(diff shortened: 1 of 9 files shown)


YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
```diff
@@ -44,69 +44,86 @@ public void OnVisitSyntaxNode(GeneratorSyntaxContext context)
...
-                            classSymbol = classSymbol.BaseType;
+                            classObject.OnSerializedMethods.Add(methodSymbol);
+                        }
+                        if (methodAttributes.Any(x => x.AttributeClass!.ToDisplayString() == "YamlDotNet.Serialization.Callbacks.OnSerializingAttribute"))
+                        {
+                            classObject.OnSerializingMethods.Add(methodSymbol);
                         }
                     }
                 }
+                classSymbol = classSymbol.BaseType;
```
EdwardCooke, line 126: This line throws a compiler warning because of the null propagation checks.

If you mark the parameter in the method signature as nullable it will get rid of the warning. You'll also need to mark the object as not null with an `!` where you get the sanitized name.

EdwardCooke:
I like this approach, this is nice. The only comment is about the null propagation compiler warning. Once that's resolved, I'll merge this in.

YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
prochnowc: As a side note: the warning was already present before ;)

prochnowc: Sorry - ignore my side note ;) The existing warning appeared on `Classes.ContainsKey()`.

prochnowc:
Glad that you like it.

When do you plan the next release including this feature ?

EdwardCooke:
In about 2 minutes, have to let the pipelines finish :)

aaubry:
This feature has been released in [version 13.5.0](https://github.com/aaubry/YamlDotNet/releases/tag/v13.5.0).

EdwardCooke:
It takes a couple of minutes for NuGet to approve the package but it will be there momentarily.

prochnowc:
Sure. Thanks for accepting the feature/PR.

prochnowc:
@EdwardCooke there is one change missing which I accidently did not commit:
https://github.com/aaubry/YamlDotNet/pull/853

PR #855, opened 2023-10-05
Fix code generation for Dictionary and List which use Array values 

### LINKED ISSUE #854: Static code generation fails to generate code for Dictionary and List which use Array values
**Describe the bug**

The static code generator throws `NotSupportedException` when using `Dictionary<string, string[]>` or `List<string, string[]>`.

**To Reproduce**

Add the source generator package to the project and add the following types:

```csharp
[YamlStaticContext]
public class MyContext { }

[YamlSerializable]
public class Model {
   public Dictionary<string, string[]> MyDictionary { get; set; }
}
```

prochnowc:
Fixes #854

### DIFF
--- a/YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
+++ b/YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
@@ -94,30 +94,40 @@ namespace YamlDotNet.Analyzers.StaticGenerator
 
                     if (member is IPropertySymbol propertySymbol)
                     {
-                        classObject.PropertySymbols.Add(propertySymbol);
-                        CheckForSupportedGeneric(propertySymbol.Type);
+                        if (!classObject.PropertySymbols.ContainsName(propertySymbol))
+                        {
+                            classObject.PropertySymbols.Add(propertySymbol);
+                            CheckForSupportedGeneric(propertySymbol.Type);
+                        }
                     }
                     else if (member is IFieldSymbol fieldSymbol)
                     {
-                        classObject.FieldSymbols.Add(fieldSymbol);
-                        CheckForSupportedGeneric(fieldSymbol.Type);
+                        if (!classObject.FieldSymbols.ContainsName(fieldSymbol))
+                        {
+                            classObject.FieldSymbols.Add(fieldSymbol);
+                            CheckForSupportedGeneric(fieldSymbol.Type);
+                        }
                     }
                     else if (member is IMethodSymbol methodSymbol)
                     {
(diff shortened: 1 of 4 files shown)


YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
```diff
@@ -94,30 +94,40 @@ private void AddSerializableClass(INamedTypeSymbol? classSymbol)
 
                     if (member is IPropertySymbol propertySymbol)
                     {
-                        classObject.PropertySymbols.Add(propertySymbol);
-                        CheckForSupportedGeneric(propertySymbol.Type);
+                        if (!classObject.PropertySymbols.ContainsName(propertySymbol))
```
prochnowc, line 97: I know these changes are a bit unrelated - but not doing the check the generator generates redundant code if the model is reachable via multiple code paths.

prochnowc:
@EdwardCooke I have another fix for static code generation prepared, so please do not do a release immediately ;)

EdwardCooke:
Sounds good. I’ll try and get to it as soon as your ready.

prochnowc:
Should I put the fix in this PR or do another one ?
It is about deserializing into `object` (which already uses Dictionary<object, object> without static code generation)

EdwardCooke:
It doesn’t always go into a dictionary. You can serialize/deserialize just a single value like an int or string. Also, you won’t know what the type of the object should be so it doesn’t know what to deserialize into which is why it throws an exception there. There was another case where it couldn’t figure it out back when I built out the static stuff and I can’t remember exactly what it is. I very vaguely remember it had something to do with dictionary object, object not working right in aot compiled applications. Might be way off there though. 

did you find a way around those object limits?

prochnowc:
At least I managed to deserialize object into Dictionary<object,object> with AOT. I will open annother issue about that, this PR doesnt seem to be the right place to discuss this. :)

EdwardCooke:
Sounds good to me. Do you want me to merge this in then?

prochnowc:
Yes please.

EdwardCooke:
Will do

YamlDotNet.Analyzers.StaticGenerator/ClassSyntaxReceiver.cs
EdwardCooke: Sounds good to me.

aaubry:
This feature has been released in [version 13.5.2](https://github.com/aaubry/YamlDotNet/releases/tag/v13.5.2).

PR #856, opened 2023-10-06
Added support for serialization and deserialization of 'object' to static code generation

prochnowc:
This PR adds support for serialization and deserialization of 'object' as long as the type is a primitive type or is known to the static context.

Changes:

* Added a `IsKnownType()` method to the `StaticContext` to determine whether a type is known to the context.
* Added code generation for a `StaticTypeResolver` which resolves types known the context.
* Modified code generation for `StaticTypeDescriptor` to use the `StaticTypeResolver` from the context.

@EdwardCooke What do you think ? If the approach is fine I will create an issue to reference this PR.

### DIFF
--- a/YamlDotNet.Test/Analyzers/StaticGenerator/RootCollectionTests.cs
+++ b/YamlDotNet.Test/Analyzers/StaticGenerator/RootCollectionTests.cs
@@ -54,6 +54,36 @@ namespace YamlDotNet.Test.Analyzers.StaticGenerator
             Assert.Equal("hello", actual[0].Test);
             Assert.Equal("world", actual[1].Test);
         }
+
+        [Fact]
+        public void RootDictionaryWorks()
+        {
+            var deserializer = new StaticDeserializerBuilder(new StaticContext()).Build();
+            var yaml = @"
+a:
+ Test: hello
+b:
+ Test: world
+";
+
+            var actual = deserializer.Deserialize<Dictionary<string, RootObject>>(yaml);
+            Assert.Equal("hello", actual["a"].Test);
+            Assert.Equal("world", actual["b"].Test);
+        }
+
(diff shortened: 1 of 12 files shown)


YamlDotNet.Test/Analyzers/StaticGenerator/RootCollectionTests.cs
```diff
@@ -54,6 +54,36 @@ public void RootListWorks()
             Assert.Equal("hello", actual[0].Test);
             Assert.Equal("world", actual[1].Test);
         }
+
+        [Fact]
+        public void RootDictionaryWorks()
+        {
```
EdwardCooke, line 60: I like these tests. Could you do one more with this yaml and instead specifying the type do it as object? Should return a value of dictionary<object, object>. Just wondering if that works.

prochnowc: Ah, got you. Added `RootNestedObjectWorks()` - and yes it works ;)

EdwardCooke: Fantastic! When I get to a computer I’m going to do another review and run a couple scenarios through it. Great work.

EdwardCooke:
Sorry it’s taking so long. Work emergency. It may not be until tomorrow I get to look at this.

EdwardCooke:
Just waiting for the build to complete and I'll get a release out

aaubry:
This feature has been released in [version 13.7.0](https://github.com/aaubry/YamlDotNet/releases/tag/v13.7.0).

prochnowc:
Thanks for having this merged. The code generation should be much more usable now.

PR #865, opened 2023-11-14
Process inline comments with DictionaryDeserializer.

evabalint:
Dear Antoine, Edward,

We need to process some data from comments in yaml files. In these comments we usually got some descriptions about data types, ranges, available values.
So I've tried to use your solution to process the yaml files, and created a CommentNodeDeserializer to process all simple values into a ValueWithComment pair and then use it in the DictionaryDeserializer. I set this behavior only if the skipComments is set to false.
I hope you can accept my solution, as we really need this feature to process our input files.

Best regards,
and many thanks,
Eva.
PS: I closed the previous pull request, because I had it on my master branch, and the appVeyor build failed when trying to get a version number.

### DIFF
--- a/YamlDotNet/Core/IParser.cs
+++ b/YamlDotNet/Core/IParser.cs
@@ -39,5 +39,15 @@ namespace YamlDotNet.Core
         /// </summary>
         /// <returns>Returns true if there are more events available, otherwise returns false.</returns>
         bool MoveNext();
+
+        /// <summary>
+        /// Skips following comment events.
+        /// </summary>
+        void SkipFollowingComments();
+
+        /// <summary>
+        /// Gets the SkipComments value from its scanner.
+        /// </summary>
+        bool SkipComments { get; }
     }
 }
--- a/YamlDotNet/Serialization/BufferedDeserialization/ParserBuffer.cs
+++ b/YamlDotNet/Serialization/BufferedDeserialization/ParserBuffer.cs
@@ -44,6 +44,7 @@ namespace YamlDotNet.Serialization.BufferedDeserialization
         /// <exception cref="ArgumentOutOfRangeException">If parser does not fit within the max depth and length specified.</exception>
         public ParserBuffer(IParser parserToBuffer, int maxDepth, int maxLength)
         {
+            SkipComments = parserToBuffer.SkipComments;
             buffer = new LinkedList<ParsingEvent>();
             buffer.AddLast(parserToBuffer.Consume<MappingStart>());
             var depth = 0;
@@ -71,6 +72,8 @@ namespace YamlDotNet.Serialization.BufferedDeserialization
         /// </summary>
         public ParsingEvent? Current => current?.Value;
 
+        public bool SkipComments { get; }
+
         /// <summary>
         /// Moves to the next event.
         /// </summary>
@@ -87,6 +90,13 @@ namespace YamlDotNet.Serialization.BufferedDeserialization
         public void Reset()
         {
             current = buffer.First;
+        }
+
+        public void SkipFollowingComments()
+        {
+            while (this.TryConsume<Comment>(out var _))
+            {
+            }
         }
     }
 }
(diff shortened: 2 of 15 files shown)


YamlDotNet/Core/IParser.cs
```diff
@@ -39,5 +39,15 @@ public interface IParser
         /// </summary>
         /// <returns>Returns true if there are more events available, otherwise returns false.</returns>
         bool MoveNext();
+
```
EdwardCooke, line 42: I'm very hesitant with adding additional fields/methods to this interface. It's implemented by other consumers of the library and I really, really badly don't want to break those. That's a major breaking change.

YamlDotNet/Serialization/NodeDeserializers/CommentNodeDeserializer.cs
```diff
@@ -0,0 +1,45 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+using System;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+
+namespace YamlDotNet.Serialization.NodeDeserializers
+{
+    internal class CommentNodeDeserializer : INodeDeserializer
```
EdwardCooke, line 28: This should be public so people can remove it if needed.

YamlDotNet/Serialization/BufferedDeserialization/ParserBuffer.cs
```diff
@@ -71,6 +72,8 @@ public ParserBuffer(IParser parserToBuffer, int maxDepth, int maxLength)
         /// </summary>
         public ParsingEvent? Current => current?.Value;
 
+        public bool SkipComments { get; }
```
EdwardCooke, line 75: This should be true by default, for sure. I'm not sure of what kind of repercussions this could have in places that inherit or use this outside of the library, I'm suspecting it could cause issues if dictionaries suddenly have additional unexpected objects.

I would like it to be disabled through these changes by default, and have to be opted into for it to work.

YamlDotNet/Serialization/DeserializerBuilder.cs
```diff
@@ -97,7 +97,8 @@ public DeserializerBuilder()
                 { typeof(DictionaryNodeDeserializer), _ => new DictionaryNodeDeserializer(objectFactory.Value, duplicateKeyChecking) },
                 { typeof(CollectionNodeDeserializer), _ => new CollectionNodeDeserializer(objectFactory.Value) },
                 { typeof(EnumerableNodeDeserializer), _ => new EnumerableNodeDeserializer() },
-                { typeof(ObjectNodeDeserializer), _ => new ObjectNodeDeserializer(objectFactory.Value, BuildTypeInspector(), ignoreUnmatched, duplicateKeyChecking, typeConverter) }
+                { typeof(CommentNodeDeserializer), _ => new CommentNodeDeserializer() },
```
EdwardCooke, line 100: This should be optional and not included by default. You'll also need to add this to the staticdeserializerbuilder for feature parity with the ahead of time compilation abilities of the library.

YamlDotNet/Core/IScanner.cs
```diff
@@ -54,5 +54,10 @@ public interface IScanner
         /// Consumes the current token.
         /// </summary>
         void ConsumeCurrent();
+
+        /// <summary>
+        /// Gets the SkipComments setting.
+        /// </summary>
+        bool SkipComments { get; }
```
EdwardCooke, line 61: Same comment as in the IParser, I'm very hesitant to modify these core interfaces.

evabalint:
Hello,
I do understand your reluctance. I'm going to restructure my code, so I don't have to modify the IScanner and IParser interfaces of your Core project, and make sure that it has no side-effects.
Many thanks for your quick reply.

PR #868, opened 2023-11-16
Process inline comments with DictionaryDeserializer-2

evabalint:
Hello again,
I've structured my modifications in separate classes under separate namespace (Core/ParsingComments/)
- ParserWithComments : Parser
- ScannerWithComments : Scanner
I've created a common base deserializer, which inherits from the base deserializer:
	- DictionaryDeserializerWithComments: DictionaryDeserializer
and I've created two dictionary deserializer classes:
- DictionaryNodeDeserializerWithComments	: DictionaryDeserializerWithComments
- StaticDictionaryNodeDeserializerWithComments	: DictionaryDeserializerWithComments
I've created a new value deserializer, called NodeCommentValueDeserializer, which does what the NodeValueDeserializer does, but handles comment-events differently. 
All these modifications can be called from the DeserializerBuilders calling its new function: BuildWithCommentsDeserializer.
I know, that I did not answer your most important question, why would anyone process comments like data, but that is how we get our input from our clients :( I'm trying to find a solution for it...
I hope this looks better :) What's your opinion?
And many thanks,

### DIFF
--- a/YamlDotNet.Samples/DeserializeWithComment.cs
+++ b/YamlDotNet.Samples/DeserializeWithComment.cs
@@ -1,0 +1,84 @@
+﻿// This file is part of YamlDotNet - A .NET library for YAML.
+// Copyright (c) Antoine Aubry and contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+// of the Software, and to permit persons to whom the Software is furnished to do
+// so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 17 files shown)


evabalint:
Hi guys, 
Do you have any thoughts about this request, please?

EdwardCooke:
I’ll try and take another look later today. When I went through it the first time it felt like a lot code duplication which isn’t great for maintaining it. Give me a few days.

EdwardCooke:
I haven’t forgotten about this.

filippobottega:
Hi @evabalint and @EdwardCooke ,
I've briefly tried to understand the PR implementation but I'm not confident that the overall design is consistent.
I suggest you to think about yaml comments as a new layer on top of the Representation Model.
Comments could be inline or block and they could be linked with the corresponding yaml node.
In the  Representation Model each node could be a dotnet object instance, a tag or a property value of an instance. Let me know if I'm wrong.
I think that we can leave the Representation Model untouched and we can create a list for each type of yaml node with comments linked to corresponding entities:

1. `List<(object, string)>` for comments of instances.
2. `List<(string, string)>` for comments of tags.
3. `List<(PropertyInfo, object, string)>` for comments of property values.

During serialization we can create these lists and pass theme to the serializer to emit comments accordingly.
During deserialization the deserializer could return comments generating the corresponding lists.
We can wrap the lists in a Comments class.

This is just a suggestion to enrich the library without changing the current behaviuor.

Thanks for your useful library,
Filippo

PR #877, opened 2023-12-24
Handle TargetInvocationException when reading a property

alexeykuptsov:
Hi,

In my project I use YamlDotNet to serialize arbitrary objects. I can't mark properties of the objects with [YamlIgnore] but I want serializer not to fail in cases when an erroneous property is met. Could you add an ability to serialize such properties as YAML strings literals with a simple error message?

### DIFF
--- a/YamlDotNet/Serialization/TypeResolvers/StaticTypeResolver.cs
+++ b/YamlDotNet/Serialization/TypeResolvers/StaticTypeResolver.cs
@@ -32,6 +32,11 @@ namespace YamlDotNet.Serialization.TypeResolvers
         {
             if (actualValue != null)
             {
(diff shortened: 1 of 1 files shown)


alexeykuptsov:
I'm not familiar with AppVeyor so I haven't found why the build has failed (by reading the build log). Maybe I should have change the package version somehow? Also didn't manage to find information in contribution documentation.

EdwardCooke:
It fails when the branch your merging from is master. I haven’t had a chance to look into why and how to fix it yet. I’m going to ask you if you can recreate this or from a branch other than master so we can have the build run correctly. I’ll look at this or when I get some time.

PR #879, opened 2023-12-26
Handle TargetInvocationException when reading a property

alexeykuptsov:
In my project I use YamlDotNet to serialize arbitrary objects. I can't mark properties of the objects with [YamlIgnore] but I want serializer not to fail in cases when an erroneous property is met. Could you add an ability to serialize such properties as YAML strings literals with a simple error message?


alexeykuptsov:
I'm sorry for this mess of closing and reopening. I'm getting used to contributing to OSS 😅

EdwardCooke:
Can also add this to the static serializer pieces of yamldotnet?

alexeykuptsov:
I've added test StaticSerializationHandlesTargetInvocationException but I don't get the static serializer output. The serializer doesn't throw and returns empty object output `{}`. Can you help me with binding field `StaticSerializerBuilder.handleTargetInvocationExceptions` with the part of code where it reads from an input object property?

EdwardCooke:
Just looked this over. One other comment, it may be better to bring the exception handling logic out of the object descriptor, instead pass in a method that would get called in the catch. Takes 3 parameters, exception, object and property. Default of that method would be throw. Then people can handle any different exceptions differently based on their needs. For example, dump out the exception message.

alexeykuptsov:
@EdwardCooke Hi Edward, I've implemented the support of custom exception handler function. But I haven't implemented default method "throw". I don't like this idea because try/catch block may affect performance in simple cases like serialization of plain objects with only auto-properties (POCO).

Issue #886, opened 2024-01-07
`Yaml.Text(...)` fumbles without trailing whitespace


dougbu:
**Describe the bug**

Many editors either automatically remove trailing whitespace or have extensions that do so. Unfortunately, when trailing whitespace is removed from (say) Serialization/SerializationTests.cs, multiple tests fail with e.g.,
``` text
[xUnit.net 00:00:03.99]     ExampleFromSpecificationIsHandledCorrectly [FAIL]
  Failed ExampleFromSpecificationIsHandledCorrectly [1 ms]
  Error Message:
   System.ArgumentOutOfRangeException : startIndex cannot be larger than length of string. (Parameter 'startIndex')
  Stack Trace:
     at System.String.Substring(Int32 startIndex, Int32 length)
   at System.String.Substring(Int32 startIndex)
   at YamlDotNet.Test.Yaml.<>c__DisplayClass10_0.<Text>b__2(String l) in C:\dd\dnx\other\YamlDotNet\YamlDotNet.Test\Yaml.cs:line 111
   at System.Linq.Enumerable.SelectListIterator`2.ToList()
   at YamlDotNet.Test.Yaml.Text(String yamlText) in C:\dd\dnx\other\YamlDotNet\YamlDotNet.Test\Yaml.cs:line 110
   at YamlDotNet.Test.Serialization.SerializationTests.ExampleFromSpecificationIsHandledCorrectly() in C:\dd\dnx\other\YamlDotNet\YamlDotNet.Test\Serialization\SerializationTests.cs:line 1230
```

That particular test fails at the empty line near the top of the `yamlText`:
``` yaml
                obj:
                  - &CENTER { x: 1, y: 2 }
                  - &LEFT { x: 0, y: 2 }
                  - &BIG { r: 10 }
                  - &SMALL { r: 1 }

                # All the following maps are equal:
```

Fix is fairly simple. I'll put up a PR in a bit&hellip;

**To Reproduce**

Edit Serialization/SerializationTests.cs and remove all trailing whitespace, at least in the `ExampleFromSpecificationIsHandledCorrectly()` fact. Then, execute that test.

dougbu:
fixed in #887 and  merged (thanks @aaubry❕) today

PR #887, opened 2024-01-07
Handle blank lines in `Yaml.Text()`

dougbu:
- see https://github.com/aaubry/YamlDotNet/issues/886
- transform blank lines to `string.Empty`
- remove all trailing whitespace from lines
  - if trailing whitespace is significant in a future test, YAML files can be added
- `throw new ArgumentException(...)` if a line is insufficiently indented
  - should be easier to fix test than with previous `ArgumentOutOfRangeException` for this case
- add tests

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -252,7 +252,7 @@ namespace YamlDotNet.Test.Serialization
         {
             var yamlReader = new StringReader(@"Text: >
   Some Text.
-  
+
 Value: foo");
             var result = Deserializer.Deserialize(yamlReader);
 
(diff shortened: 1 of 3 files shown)


YamlDotNet.Test/Serialization/SerializationTests.cs
```diff
@@ -252,7 +252,7 @@ public void DeserializeWithGapsBetweenKeys()
         {
             var yamlReader = new StringReader(@"Text: >
   Some Text.
-  
+
```
dougbu, line 255: whitespace-only changes in this file are essentially end-to-end tests of `Yaml.Text()` :smile:

YamlDotNet.Test/YamlTests.cs
```diff
@@ -0,0 +1,116 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+using System;
+using FluentAssertions;
+using Xunit;
+
+namespace YamlDotNet.Test
+{
+    public class YamlTests
```
dougbu, line 28: no, this test class does not cover the `Yaml` test helper class comprehensively. it covers what I changed in `Yaml.Text()` however

aaubry:

Issue #807, opened 2023-05-19
Set anchor tag name during serialization using YamlDotNet?


Nidhi1819:
This sounds interesting. Is there also an easy way to control the anchor naming during serialization? I would like to use YamlDotNet to create a stable file format, that changes as little as possible during multiple deserialize, change, serialize cycles, so different versions of the files can be compared and merged easily.

_Originally posted by @rohahn in https://github.com/aaubry/YamlDotNet/issues/227#issuecomment-311423294_

EdwardCooke:
I think you can replace this implementation with your own and be able to create more meaningful names. However, you may not be able to get meaningful context to generate a name that makes sense. 

https://github.com/aaubry/YamlDotNet/blob/9a5900fb35e5cb668a77d2449b96f8bc85c179ba/YamlDotNet/Serialization/ObjectGraphVisitors/AnchorAssigner.cs#L29

EdwardCooke:
Since there was a solution provided with no feedback I’m closing this issue. If you disagree go ahead and reopen it and we can discuss more.

BBoldenow-Alt:
> Since there was a solution provided with no feedback I’m closing this issue. If you disagree go ahead and reopen it and we can discuss more.

@EdwardCooke - I think this is reasonable, but am having trouble getting the IAliasProvider implementation to trigger. I mostly copied the AnchorAssigner, but customized the name portion. I then provided this to my serializer like this:

```csharp
var anchorNaming = new CustomAnchorAssigner(new List<IYamlTypeConverter>()
{
    customConverter1,
    customConverter2,
    customConverter3,
});

var serializerBuilder = new SerializerBuilder()
    .ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitEmptyCollections)
    .ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitNull)
    .WithTypeConverter(customConverter1)
    .WithTypeConverter(customConverter2)
    .WithTypeConverter(customConverter3)
    .WithPreProcessingPhaseObjectGraphVisitor(anchorNaming)
```

This triggers all of the `VisitObject` events, but the `GetAlias` method is never called. Is there something I'm missing to use this Alias Provider?

Schrolli91:
@BBoldenow-Alt @EdwardCooke 
I am a little bit confused, how to implement that.
Is there a little more complete example, how to get a custom anchor naming working?

EdwardCooke:
[...]

PR #895, opened 2024-01-21
Make the type converters available to custom pre-processing phase visitors.

### LINKED ISSUE #677: How to get type converters when using `WithPreProcessingPhaseObjectGraphVisitor`
I'm currently trying to solve #660 in our project, by duplicating `AnchorAssigner` and replacing `assignments` dictionary definition to a compare by reference one. However, when I'm building the serializer, I found that `WithPreProcessingPhaseObjectGraphVisitor()` doesn't take a `Func` type as the parameter, so I need to instantiate my own `ReferenceEqualAnchorAssigner` but I don't know how can I get type converters.

So I would like to know is there a way to get a list of type converters?

My code is

```cs
new SerializerBuilder()
    .ConfigureDefaultValuesHandling(DefaultValuesHandling.Preserve)
    // expecting something `like typeConverters => new ReferenceEqualAnchorAssigner(typeConverters)` for the first argument
    .WithPreProcessingPhaseObjectGraphVisitor(new ReferenceEqualAnchorAssigner( ??? ), w => w.InsteadOf<AnchorAssigner>())
    .WithEmissionPhaseObjectGraphVisitor(args => new AnchorAssigningObjectGraphVisitor(args.InnerVisitor, args.EventEmitter, args.GetPreProcessingPhaseObjectGraphVisitor<ReferenceEqualAnchorAssigner>()),
        w => w.InsteadOf<AnchorAssigningObjectGraphVisitor>())
    .Build();
```

and `ReferenceEqualAnchorAssigner` is the same as `AnchorAssigner` except `assignments`:

```cs
public sealed class ReferenceEqualAnchorAssigner : PreProcessingPhaseObjectGraphVisitorSkeleton, IAliasProvider {
    private readonly IDictionary<object, AnchorAssignment> assignments = new Dictionary<object, AnchorAssignment>(ReferenceEqualityComparer.Instance);
    ...
}
```

stijnherreman:
This change makes the type converters available to custom pre-processing phase visitors instead of discarding them. The type converters are necessary when inheriting from `PreProcessingPhaseObjectGraphVisitorSkeleton`.

The change also aligns the `WithPreProcessingPhaseObjectGraphVisitor(...)` API with the `WithEmissionPhaseObjectGraphVisitor(...)` API.

Fixes #677 

There are currently no unit tests at all for this API, but I can add one if required. I'm already using this change locally to implement a custom anchor assigning visitor (see #807).

### DIFF
--- a/YamlDotNet/Serialization/SerializerBuilder.cs
+++ b/YamlDotNet/Serialization/SerializerBuilder.cs
@@ -377,6 +377,22 @@ namespace YamlDotNet.Serialization
         /// before actually emitting it. This allows a visitor to collect information about the graph that
         /// can be used later by another visitor registered in the emission phase.
         /// </remarks>
+        /// <param name="objectGraphVisitorFactory">A function that instantiates the type inspector.</param>
+        public SerializerBuilder WithPreProcessingPhaseObjectGraphVisitor<TObjectGraphVisitor>(Func<IEnumerable<IYamlTypeConverter>, TObjectGraphVisitor> objectGraphVisitorFactory)
+            where TObjectGraphVisitor : IObjectGraphVisitor<Nothing>
+        {
+            return WithPreProcessingPhaseObjectGraphVisitor(objectGraphVisitorFactory, w => w.OnTop());
+        }
+
+        /// <summary>
+        /// Registers an additional <see cref="IObjectGraphVisitor{Nothing}" /> to be used by the serializer
+        /// before emitting an object graph.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


EdwardCooke:
I'm fine pushing in another breaking change in the next version, there's already going to be a couple. However, can you just add method overloads to incorporate your changes?

stijnherreman:
Sure, either is fine with me. I've updated the PR to add overloads instead of replacing the existing ones.

aaubry:

Issue #773, opened 2023-01-30
Support naming conventions for values


pardahlman:
**Is your feature request related to a problem? Please describe.**

I have a YAML file where both properties and their values use snake case.

```yaml
project:
    type: database_migration
```

The corresponding C# DTO uses enum

```csharp
enum ProjectType
{
  DatabaseMigration
}
```

I have configred YamlDotNet to use `UnderscoredNamingConvention`, but the deserialization fails:

```
YamlDotNet.Core.YamlException: (Line: 4, Col: 11, Idx: 36) - (Line: 4, Col: 20, Idx: 45): Exception during deserialization
 ---> System.ArgumentException: Requested value 'database_migration' was not found.
   at System.Enum.TryParseByName(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, UInt64& result)
   at System.Enum.TryParseInt32Enum(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Int32 minInclusive, Int32 maxInclusive, Boolean ignoreCase, Boolean throwOnFailure, TypeCode type, Int32& result)
   at System.Enum.TryParse(Type enumType, String value, Boolean ignoreCase, Boolean throwOnFailure, Object& result)
   at System.Enum.Parse(Type enumType, String value, Boolean ignoreCase)
   at YamlDotNet.Serialization.NodeDeserializers.ScalarNodeDeserializer.YamlDotNet.Serialization.INodeDeserializer.Deserialize(IParser parser, Type expectedType, Func`3 nestedObjectDeserializer, Object& value)
```

My initial analysis suggests that it's the call to `Enum.Parse` in [ScalarNodeDeserializer](https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs#L62) that is performed with the underscored value (`database_migration`). Looking further into how `INamingConvention` is used, it appears that it's only used for property values (I might be wrong here!).

**Describe the solution you'd like**

I would like to deserialize YAML with snake case property values into C# enums defined in pascal case. It would be preferable if I could serialize C# DTOs using pascal case enums values into YAML with snake case property values (e.g. roundtrip).

**Describe alternatives you've considered**

Using a C# enum with snake case works, but is not desirable

```csharp
enum ProjectType
{
  database_migration
}
```

EdwardCooke:
[...]

PR #897, opened 2024-01-22
Ec net80 enumsasnumbers

EdwardCooke:
Enables #773 

Adds naming conventions for convert enums to/from scalars
Adds .NET8 binaries
Adds support for custom formatting of enums, easy to expose enums as integers or whatever else.
Quotes necessary enum strings (Null for example)

**Breaking: For those that get impacted pass in `NullNamingConvention.Instance` to the `EnumNamingConvetion` arguments on the constructor**
**Breaking: Removed many of the redundant constructors for the classes, pass in the old default values to the new constructors**


aaubry:

PR #902, opened 2024-02-04
Fixed #901 - private constructor and tag issue

### LINKED ISSUE #901: 15.1.0 EnsureRoundTrip no longer works without tag mapping and protected default constructors
**Describe the bug**
When trying out the new release I encountered two problems with my project.
1. `EnsureRoundtrip()` does not work if I use a tag mapping for each class, but it worked before. Why do I need a tag mapping if the type can be resolved through the object graph?
2. `EnablePrivateConstructors()` used to enable protected constructors as well. Now I have to use private or public constructors

**To Reproduce**
```
public void SerializeTestClass()
{
    var testClass = new TestClass("test");
    var serializer = new SerializerBuilder()
        .EnablePrivateConstructors()
        .IncludeNonPublicProperties()
        .EnsureRoundtrip()
        .Build();
    
        .EnablePrivateConstructors()
        .IncludeNonPublicProperties()
        .WithNamingConvention(PascalCaseNamingConvention.Instance)
        .Build();

    var stringWriter = new StringWriter();
    serializer.Serialize(stringWriter, testClass);
  }

 public class TestClass
 {
    public TestClass()
    {
        // protected constructor
    }

    public TestClass(string value)
    {
        Value = value;
    }

    public string Value { get; private set; }
}
```

This works in 13.7.1, but not in 15.1.0:
```
Cannot serialize type 'TestClass' where a 'System.Object' was expected because no tag mapping has been registered for TestClass', which means that it won't be possible to deserialize the document.
```
` .WithTagMapping(new TagName("!Test"), typeof(TestClass))` does fix the problem, but that doesn't help for existing yaml files.

These are very breaking changes and make backwards compability impossible. For each class I serialize, I'd have to add a type mapping and make the constructor private instead of protected. Is there any solution I don't know about?

EdwardCooke:
Fixes #901

### DIFF
--- a/YamlDotNet.Test/Serialization/PrivateConstructorTests.cs
+++ b/YamlDotNet.Test/Serialization/PrivateConstructorTests.cs
@@ -20,6 +20,7 @@
 // SOFTWARE.
 
 using System;
+using System.IO;
 using Xunit;
 using YamlDotNet.Core;
 using YamlDotNet.Serialization;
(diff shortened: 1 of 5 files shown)


aaubry:
This feature has been released in [version 15.1.1](https://github.com/aaubry/YamlDotNet/releases/tag/v15.1.1).

Issue #893, opened 2024-01-17
JsonCompatible serializer is not Json compatible


Rackover:
**Describe the bug**
When serializing an object that contains an infinity float, the YAML DotNet serializer in JsonCompatible mode produces a JSON that contains the keyword `.inf` . 
According to JSON spec it should write `null`, otherwise i suppose writing a very big number or enclosing `.inf` with quotes would be appreciated. Currently the JSON produced by the serializer in that context is unusable for other parsers that expect valid data.

**To Reproduce**
```
 YamlDotNet.Serialization.SerializerBuilder serializerBuilder = new();
 serializerBuilder.JsonCompatible();
 serializerBuilder.WithIndentedSequences();

 var serializer = serializerBuilder.Build();

class Test
{
float a = float.PositiveInfinity;
}

var str = serializer.Serialize(a);
Console.Write(str);
```

EdwardCooke:
Thanks for the report. It’s on my todo list. Might be a bit before I get to it. If you need it sooner go ahead and create a pr. You’ll need to handle other data types like double as well.

Rackover:
Might have invented it, it seems the real status of Infinity and NaN is unclear:
https://www.ietf.org/rfc/rfc4627.txt

So it could be Null, it could be Undefined, or as someone proposed here https://stackoverflow.com/a/28763430/6230450 it could be the string "Infinity". Either choice is good I suppose as long as the final produced JSON is valid to deserialize.

Rackover:
```
            string serialized = serializer.Serialize(heatpoint);
            Regex reg = new Regex("(\":) (\\.inf)");
            serialized = reg.Replace(serialized, "$1 0");
```

Currently using this as a workaround whenever i do any JSON serialization with the YAMLDotNet serializer (in case this helps anyone)

EdwardCooke:
This will be fixed in the next release.

PR #907, opened 2024-02-19
Enquote value in JSonEmitter when it is an infinity float or double

Rackover:
Stub for fixing https://github.com/aaubry/YamlDotNet/issues/893
Incomplete - missing unit test, and this might not be the best way to handle it either (i haven't looked too much into it). but I guess it's a start?


EdwardCooke:
Would you add the unit tests for this? Otherwise I can take it and add them in in the next week or 2.

Rackover:
This PR looks like it is now obsolete :)
Looking at the merge conflict over here

PR #912, opened 2024-02-26
Fixed #732, quote strings starting with whitespace characters

### LINKED ISSUE #732: String starting with tab and comma is incorrectly serialized leading to SemanticErrorException when that yaml is deserialized
The following code is the full code to reproduce the bug. (Its commented.)

var serializer = new SerializerBuilder().Build();
var yaml = serializer.Serialize(new List<string>() { "\t," }); // Fails to decorate the yaml string, which causes the exception when deserializing - see below
deserializer.Deserialize<List<string>>(yaml); // SemanticErrorException : (Line: 1, Col: 4, Idx: 3) - (Line: 1, Col: 5, Idx: 4): While parsing a node, did not find expected node 

Project settings
C#
<TargetFramework>net6.0-windows10.0.22000.0</TargetFramework>
<PackageReference Include="YamlDotNet" Version="12.0.2" />

Thanks

EdwardCooke:
Fixes #732

### DIFF
--- a/YamlDotNet.Benchmark/Program.cs
+++ b/YamlDotNet.Benchmark/Program.cs
@@ -1,4 +1,11 @@
-﻿using BenchmarkDotNet.Running;
-using YamlDotNet.Benchmark;
+﻿using YamlDotNet.Serialization;
 
-BenchmarkSwitcher.FromAssembly(typeof(YamlStreamBenchmark).Assembly).Run(args);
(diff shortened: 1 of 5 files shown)


aaubry:
This feature has been released in [version 15.1.2](https://github.com/aaubry/YamlDotNet/releases/tag/v15.1.2).

PR #924, opened 2024-05-14
use correct framework

kasperk81:
the undocumented names (without period in dotnet 5+) can drop their support when move to double-digit (to avoid potential conflicts) https://learn.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks


aaubry:
This feature has been released in [version 15.1.6](https://github.com/aaubry/YamlDotNet/releases/tag/v15.1.6).

PR #927, opened 2024-06-02
Optimize YamlScalarNode.Load and YamlMappingNode.Load

lahma:
Improving scalar loading performance a bit as it's done a lot for common files. Check for "null" marker faster and don't do try-catch in a loop. I fixed the saltern example YAML file for easier testing.


aaubry:
This feature has been released in [version 16.0.0](https://github.com/aaubry/YamlDotNet/releases/tag/v16.0.0).

PR #928, opened 2024-06-04
Enable .NET analyzers 

lahma:
* enabled for main project, added exclusion for things that would make changes bigger
* fix where makes sense 
* added couple polyfills that allow using better APIs on newer TFMs

Currently everything else is just a warning and allows gradually fix things when wanted in smaller PRs.


EdwardCooke:
Well want to get rid of the compiler warnings before merging this in.  Other than that it looks good. I’ll get it in once those are fixed so it goes out in the next release.

lahma:
Added the missing `nameof` usage, no warnings in build output.

PR #930, opened 2024-06-15
Add FSharpOption support

### LINKED ISSUE #913: FSharp deserialization of records with optional record fields
I'm trying to figure out the canonical way to deserialize a record with an optional field that is not a scalar value.

Given the script:

```fsharp
[<CLIMutable>]
type SubThing = { value: string }

[<CLIMutable>]
type Thing = { subThing: SubThing option }

open YamlDotNet.Serialization

let d = Deserializer()

"""
subThing:
  value: "a"
"""
|> d.Deserialize<Thing>
|> printfn "%A"
```

Running this fails with the cryptic error:

```
(Line: 3, Col: 3, Idx: 13) - (Line: 3, Col: 3, Idx: 13): Exception during deserialization
Stopped due to error
```

I would expect this to result in `{ subThing = Some { value = "a" } }`.

A simple omission of the field succeeds in the way I would expect:

```fsharp
"""
{}
"""
|> d.Deserialize<Thing>
|> printfn "%A"
```

```
{ subThing = None }
```

However, if the type of the `subThing` field is a scalar (e.g. `string`), the behavior is as I expect; omitting it results in `None` and supplying it results in `Some <value>`. 

Is this behavior configurable in some way?

MrLuje:
Fixes #913

Add support for serialization/deserialization of FSharpOption<> type

I may have missed some usecases, but here's what I though of:
- scalar & objects types (regular enumerable should work OOB from what I see)
- DefaultValue is not really a thing for F# record, so out of scope
- DefaultValuesHandling works

### DIFF
--- a/YamlDotNet.Fsharp.Test/DeserializerTests.fs
+++ b/YamlDotNet.Fsharp.Test/DeserializerTests.fs
@@ -1,0 +1,87 @@
+module DeserializerTests
+
+open System
+open Xunit
+open YamlDotNet.Serialization
+open YamlDotNet.Serialization.NamingConventions
+open FsUnit.Xunit
+open System.ComponentModel
+
+[<CLIMutable>]
+type Spec = {
+    EngineType: string
+    DriveType: string
+}
+
+[<CLIMutable>]
+type Car = {
+    Name: string
+    Year: int
+    Spec: Spec option
+    Nickname: string option
+}
+
+[<CLIMutable>]
+type Person = {
+    Name: string
+    MomentOfBirth: DateTime
+    Cars: Car array
+}
+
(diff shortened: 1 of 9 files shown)


aaubry:
This feature has been released in [version 15.3.0](https://github.com/aaubry/YamlDotNet/releases/tag/v15.3.0).

AntyaDev:
Hi @MrLuje 

Thank you for your PR. I tried this type, and it fails with `Dictionary<string, string> option`.
I used YamlDotnet 17.1.0

```
[<CLIMutable>]
type ContainerSpec = {
    Resources: string
    NodeSelector: Dictionary<string, string> option
}
```

PR #937, opened 2024-06-19
ArrayNodeDeserializer can now deserialize circular references.

### LINKED ISSUE #933: Aliases do not resolve correctly in Sequences
https://github.com/aaubry/YamlDotNet/blob/23991bd3d22009f7478e9e5c3ac71b106a8da3b7/YamlDotNet/Serialization/NodeDeserializers/ArrayNodeDeserializer.cs#L48C13-L53C1

The `ArrayNodeDeserializer` leverages `CollectionNodeDeserializer.DeserializeHelper` to deserialize nodes, however, because it doesn't know the final size of the array being deserialized it uses an `ArrayList` as a temporary result bucket.  Because the `IList result` it passed into `CollectionNodeDeserializer.DeserializeHelper` is not the final array, any `ValuePromises` that get eventually resolved [here](https://github.com/aaubry/YamlDotNet/blob/23991bd3d22009f7478e9e5c3ac71b106a8da3b7/YamlDotNet/Serialization/NodeDeserializers/CollectionNodeDeserializer.cs#L95C56-L95C129) are not reflected in the final array.

I have been able to get around this by using the attached replacement [AnchorSafeArrayNodeDeserializer.cs](https://github.com/user-attachments/files/15864238/AnchorSafeArrayNodeDeserializer.cs.txt).

You will need to configure your deserialisation builder as follows to use it:

```
var namingConvention = YamlDotNet.Serialization.NamingConventions.NullNamingConvention.Instance;

var builder =
    new DeserializerBuilder()
        .WithNamingConvention(namingConvention)
        .WithNodeDeserializer(
            inner => new AnchorSafeArrayNodeDeserializer(namingConvention),
            s => s.InsteadOf<ArrayNodeDeserializer>());
```

It should be noted that the provided `AnchorSafeArrayNodeDeserializer` uses the BCL `ArrayList` rather than the version originally defined inline.  As such, this solution will probably need to be adjusted if used in a PR.

I would create a PR myself, but I just don't have the time at the moment.  I hope this still helps people.

If it helps, I believe the following yaml should reproduce the original issue:
```yaml
items:
- &g000_FirstGroup
  name: FirstGroup
- &g001_SecondGroup
  name: SecondGroup
  child: &c000_FirstChild
    name: FirstChild
    groups:
    - *g001_SecondGroup
    - *g000_FirstGroup
    - &g002_ThirdGroup
      name: thirdGroup
```

When deserialized the `obj.items[1].groups` array will have a `null` first entry.

MetaFight:
Resolves #933 

The `ArrayNodeDeserializer` delegates the deserialization of elements to the `CollectionNodeDeserializer.DeserializeHelper`.  This, in turn, handles circular references by relying on the `IValuePromise.ValueAvailable` event to write all the resolved references to its `IList result` parameter.

However, because `ArrayNodeDeserializer` doesn't know the size of its resulting array beforehand, it uses a temporary `ArrayList` which it passes to `CollectionNodeDeserializer.DeserializeHelper`.  As a result, all resolved ValuePromise values are written to the temporary `ArrayList` instead of the final `Array`.

This PR fixes this by adding an optional `Action<int, object?>? promiseResolvedHandler` to `CollectionNodeDeserializer.DeserializeHelper`.  When provided, it is used when ValuePromises are resolved.  Otherwise, the existing behaviour is preserved.

### DIFF
--- a/YamlDotNet.Test/Serialization/DeserializerTest.cs
+++ b/YamlDotNet.Test/Serialization/DeserializerTest.cs
@@ -111,6 +111,36 @@ cars:
         }
 
         [Fact]
+        public void Deserialize_YamlWithCircularReferenceInArray_ReturnsModel()
+        {
+            var yaml = @"
+sessions:
+- &s001_FunWithLetters
+  name: Fun With Letters
+  greeter: &a001_Jill
+    name: Jill
+    sessions:
+    - *s001_FunWithLetters
+attendees:
+- *a001_Jill
+";
+
+            var sut = new DeserializerBuilder()
+                .WithNamingConvention(CamelCaseNamingConvention.Instance)
+                .Build();
+
+            var conference = sut.Deserialize<Conference>(yaml);
+            var jill = conference.Attendees.Single(a => a.Name == "Jill");
+            var session = conference.Sessions.Single(s => s.Name == "Fun With Letters");
+
+            jill.Should().NotBeNull();
+            jill.Sessions.ShouldBeEquivalentTo(new[] { session });
+
+            session.Should().NotBeNull();
+            session.Greeter.Should().Be(jill);
+        }
+
+        [Fact]
         public void SetterOnlySetsWithoutException()
         {
             var yaml = @"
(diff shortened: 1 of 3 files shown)


MetaFight:
I can't figure out why git thinks the entirety of the unit test file has changed.  I've tried switching the line endings but I think in doing that (amending the commit and using `push --force`) I broke the CI build.

I'll have some time tomorrow to try and fix this.

EdwardCooke:
The reason the CI failed is your branch name, when the PR branch is master the build fails.

MetaFight:
I'm abandoning this PR because of the branch name issue.  I'll create a new one with a better branch name.

MetaFight:
I'm abandoning this PR because of the branch name issue. I'll create a new one with a better branch name.

Issue #890, opened 2024-01-08
Questions about Custom Serialization


cd-zhang2020:
Hello，Is it possible to add attributes to specified fields to implement a custom serialization and deserialization method?

For example   Newtonsoft.Json

[JsonConverter(typeof(CustomConverter))]

` var serializer = new SerializerBuilder()
                .WithTypeConverter(new CustomConverter())
                .Build();`

I use the above code to customize the serialization, but this is a global method

EdwardCooke:
Not that I’m aware of. You’d need to have multiple builders to do different type converters like your asking about.

cd-zhang2020:
@EdwardCooke Thank you for your answer，The functionality I want to implement cannot be handled with different types converters。
For example
` serverConnector:
   #Descirption: id of server connector.
   id: server1
   addresses:
     - 127.0.0.1:8091
   protocol: grpc
   connectTimeout: 500ms
   messageTimeout: 5s
   connectionIdleTimeout: 60s
   serverSwitchInterval: 10m
   reconnectInterval: 500ms`

Deserialize the string 500ms into the number 500, and deserialize the string 5s into the number 5000

cd-zhang2020:
@EdwardCooke Thank you for your answer，The functionality I want to implement cannot be handled with different types converters。

For example 
```
serverConnector:
   id: server1
   addresses:
     - 127.0.0.1:8091
   protocol: grpc
   connectTimeout: 500ms
   messageTimeout: 5s
   connectionIdleTimeout: 60s
   serverSwitchInterval: 10m
   reconnectInterval: 500ms
``` 

Deserialize the string 500ms into the number 500, and deserialize the string 5s into the number 5000，I can't convert all strings to numbers

EdwardCooke:
I’ll work on adding this feature.

EdwardCooke:
You can now specify a custom type converter on the property using the `YamlConverter` attribute. Be sure to register the converter with the serializer builder so it's used. In the AcceptsType method, just return false.

Here's the tests that show how to do it.

https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet.Test/Serialization/TypeConverterAttributeTests.cs

Issue #884, opened 2024-01-02
IEnumerable support in StaticGenerator


tg123:
**Is your feature request related to a problem? Please describe.**

```
    // public List<string>? MyList { get; set; }
    public IEnumerable<string>? MyList { get; set; }
```

allow `IEnumerable`/`IReadOnlyList`/`IList` in StaticGenerator

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

EdwardCooke:
In addition to those interfaces, I’m adding collection and ireadonlydictionary. The code is complete and working. Just need stable WiFi.

EdwardCooke:
This is released and working now.

Issue #781, opened 2023-02-20
Could you please think about offering a `CaseInsensitiveNamingConvention`?


schittli:
Hello

**thank you very much for sharing your great work!**
I use YamDotNet for non-profit projects and it's great to have your library.

### Is your feature request related to a problem? Please describe.

Our technically quite skilled users can edit the yaml files directly.
By far the most common mistakes they make are in the capitalization of property names.
Therefore, they either need support to fix the issue or we have to program the application fault-tolerant.
But because the property names are 99% correct and only the case is wrong, there is a much more elegant solution… 

### Describe the solution you'd like

It would be very helpful if you could provide this `INamingConvention` provider: `CaseInsensitiveNamingConvention`
This would allow us to easily make YamlDotNet more fault tolerant 😊

### Describe alternatives you've considered
Of course, any application you could somehow handle these errors itself.

**But if the application is allowed to be fault tolerant with the configuration files**, 
then something like a `CaseInsensitiveNamingConvention` would be a very elegant solution.

Thanks a lot, kind regards,
Thomas

EdwardCooke:
You’ll see this feature in the next release.

EdwardCooke:
This is released

Issue #763, opened 2022-12-31
YamlDotNet doesn't respect non-nullable types


RobotsOnDrugs:
**Describe the bug**
If the YAML has a value set to `null`, but the object does not declare the type as nullable, YamlDotNet forces it to be `null` anyway (reference types) or the default value (value types). This can lead to strange unexpected behavior (especially for value types) since the application has no way of knowing whether the default value was set or if it was set to null. It should throw a YamlException just as it would for any other wrong type (e.g., deserializing "hello" to a boolean).

**To Reproduce**
Define an object with a non-nullable property and deserialize to it with YAML where the property is set to `null`.

RobotsOnDrugs:
```cs
public class MyClass
{
  public string SomeString { get; set; }
  public string? SomeNullableString { get; set; }
  public bool IsGood { get; set; }
  public bool IsNotGood { get; set; }
  public bool IsDefinitelyNotGood { get; set; }
}
```
```yaml
SomeString: null
SomeNullableString: null
IsGood: true
IsNotGood: null
IsDefinitelyNotGood: hello
```
`SomeString` should not be null, and this yaml should cause an exception at deserialization because `null` is not valid for that property and setting it to null at compile time will result in an error, yet it gets set to `null` anyway.
`SomeNullableString` can be null, so this is fine. IsGood is `true`, so that's fine as well.
`IsNotGood` should also cause an exception because `null` is not valid here. It will get set to `false` even though the yaml has a different value. This is pretty problematic for the reasons I mentioned before.
`IsDefinitelyNotGood` should cause an exception because "hello" is not valid, but YamlDotNet *will* throw an exception in this case. This is inconsistent - `null` is no more valid than "hello" for IsNotGood, yet it is set to `false` in this case.

I *could* use `Nullable<bool>`, but `null` is not valid here. The code would communicate to a reader that `null` is valid, and there would be a need to check for `null` later on, the avoidance of which is a major reason the concept of nullability was implemented in the first place. True that this is not a strict requirement at runtime, but it goes against what the compiler enforces and what the code communicates. It's perfectly valid to initialize a non-nullable type to `null`, but marking it with a `!` communicates that the value has no default and that it really needs to be set before it is used - the program *is supposed* have an error if it is not (especially when paired with the use of `init`).

Given that it is a change in behavior and that older versions of .NET would consider types not denoted with `?` as nullable, I do agree that making it opt-in is perfectly reasonable, and I appreciate that you're willing to looking into it.

[...]

PR #941, opened 2024-07-14
Fix bugs and add features

### LINKED ISSUE #845: DateTime fails Serialization/Deserialization roundtrip when JsonCompatible is on
Hi!

The following test case appears to fail:
```c#
using System.Diagnostics;

public class Program
{
    public static void Main(string[] argv)
    {
        var ser = new SerializerBuilder().JsonCompatible().Build();
        var des = new DeserializerBuilder().Build();

        var src = new DateTime(656788608000000000L, DateTimeKind.Unspecified);
        var yaml = ser.Serialize(src);
        var dst = des.Deserialize<DateTime>(yaml);

        Debug.Assert(src == dst, $"{src} vs {dst}");
    }
}
```
Values seem to differ by my local timezone offset, so if your timezone is UTC then it may work, I guess? It also seems to work correctly with `DateTimeKind.Utc` or when `JsonCompatible()` is not present.

Tested with:
```
<PackageReference Include="YamlDotNet" Version="13.3.1" />
```

### LINKED ISSUE #891: Invalid date format used in serialized JSON
**Describe the bug**

Imagine the following YAML..

```
Example date: 2023-10-12
```

.. is deserialized into an object defined as follows:

```
    public class ExampleClass
    {
        [YamlMember(Alias = "Example date")]
        public DateOnly ExampleDate { get; set; }
    }
```

With an instance of the above instantiated as "myExample", we convert to JSON as follows:

```
var serializer = new SerializerBuilder().JsonCompatible().Build();
var json = serializer.Serialize(myExample);
```

Result:

```
{"Example date": "10/12/2023"}
```

The above result is invalid because the date is formatted incorrectly.

The resultant JSON should pass validation against the following JSON schema:

```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "Example date": {
      "type": "string",
      "format": "date"
    }
  },
  "required": ["Example date"]
}
```

As such, the following result, which passes validation, is expected:

```
{"Example date":"2023-10-12"}
```

**To Reproduce**
To reproduce, simply follow the above steps and observe the output's validation result.

I used the following site to validate the JSON in this example:

https://www.jsonschemavalidator.net/

### LINKED ISSUE #893: JsonCompatible serializer is not Json compatible

### LINKED ISSUE #922: Very large files fail to parse
**Describe the bug**

I'm currently trying to parse a very large YAML file, approximately 4.8 GB, and it's failing.

**UPDATE:** I was able to parse the 4.8 GB file by changing the type of `Index` from `int` to `long` on `Mark`, `Cursor`, and `SimpleKey`.

**To Reproduce**

Try to parse a YAML file that is more than ~2.2 GB in size.

EdwardCooke:
Fix #922 - large files fail to parse
Fix #893 - JSON Compatible and non numeric numbers (.inf/nan/e+)
**BREAKING** Fix #891 - Use ISO8601 for Json Dates by default.
**BREAKING** Feat: #890 - Support custom converters by attributes
Feat: #884 - IEnumerable and other interface support in AoT
Fix #845 - DateTime/Json Compatible
Feat: #781 - Case insensitive field and property matching
Feat #763 - Non nullable properties/fields throw exception when enabled.
Feat #763 - Required fields throw exception when not set when enabled.
Feat #586 - EnumMember attribute on enums
Additionally
**BREAKING** - Support easily serializing/deserializing objects in IYamlTypeConverter implementations.
Removed the .net 7 target

----

Notes:
* The `ITypeConverter` object is now passed to many methods and is expected when some are called. You can get the typeconverter by calling `BuildTypeConverter` on the serializerbuilder and deserializerbuilder. This typeconverter should be treated as a singleton and injected into the constructor of your class.

* The `IPropertyDescriptor` is now passed into a lot of methods. It can be safely disregarded if you don't need it. It contains information about the property containing the object being deserialized.

* The delegates `ObjectSerializer` and `ObjectDeserializer` are also being passed around. They can be disregarded if you don't need them.

* To revert to the old date/time formats for `JsonCompatable` remove the `DateTime8601Converter` and add the `DateTimeConverter` with the `doubleQuotes` constructor parameter set to `true`. Example:
    ```csharp
    new SerializerBuilder()
        .WithoutTypeConverter<YamlDotNet.Serialization.Converters.DateTime8601Converter>()
        .WithTypeConverter(new YamlDotNet.Serialization.Converters.DateTimeConverter(doubleQuotes: true))
    .Build();
    ```


aaubry:
This feature has been released in [version 16.0.0](https://github.com/aaubry/YamlDotNet/releases/tag/v16.0.0).

EdwardCooke:
Release notes updated

Issue #943, opened 2024-07-25
Deserialization into generic dictionary using WithAttemptingUnquotedStringTypeDeserialization deserializes hexadecimal value incorrectly.


john-glover:
**Describe the bug**
Having built a deserializer using the `WithAttemptingUnquotedStringTypeDeserialization` option and using generic `Dictionary<object, object>` deserialization, hexadecimal values can be converted into a negative value inappropriately.

**To Reproduce**
The issue can be clearly seen in this .Net Fiddle: https://dotnetfiddle.net/Y5PPJq. 

The simplest steps are to use a value between 0x8000 and 0xFFFF (this is the range between 2 ^15 + 1 and 2 ^ 16). These values will successfully `Convert.ToInt16()` but the code makes a decision to interpret the first bit as the sign. The same thing happens at `Int32.MaxValue + 1`.

**Resolution**
I would be glad to contribute to the solution, but I don't want to spend time to create a pull request if it's not in keeping with your intent. I feel the appropriate solution is to simply use UInt64 rather than trying to use the smallest possible type -- this handles the most common cases, is not ambiguous (because yaml hex numbers have no concept of how wide they are).

**Workaround**
In my use case I've worked around the issue by writing an `INodeDeserializer` explicitly for hexadecimal values and using the `.WithNodeDeserializer` method register it instead of (wrapping) the `ScalarNodeDeserializer`. 

**Extra Bonus Bug**
The octal implementation is entirely incorrect as well, but for a different reason. The `Convert.ToInt16` (and other methods) won't convert a string with the `0o` prefix, it has to be numbers matching the regular expression `[0-8]+`. The regular expression for octal is copied straight from the hex code (which is a third issue, but entirely hidden by the complete failure to parse with the `0o`).

PR #947, opened 2024-08-03
Fix for issue 943: improper parsing of hexadecimal and octal scalars

john-glover:
This is a fix for [issue 943](https://github.com/aaubry/YamlDotNet/issues/943) that I reported. The root of the issue is that parsing sometimes surprisingly resulted in negative numbers and there was no way to specify some positive values with hexadecimal or octal scalars. The problem stems from trying to guess the width of the value because YAML does not provide a means of doing this (see [issue 943](https://github.com/aaubry/YamlDotNet/issues/943) for details).

This changes the behavior for hexadecimal deserialization so I'm prepared for there to be some controversy here, but I think this is more in line with how a developer using the library might expect the deserialization to behave. 

This also fixes the issue with octal parsing which was not functioning properly before.

### DIFF
--- a/YamlDotNet.Test/Serialization/SerializationTests.cs
+++ b/YamlDotNet.Test/Serialization/SerializationTests.cs
@@ -174,9 +174,7 @@ namespace YamlDotNet.Test.Serialization
                 .WithAttemptingUnquotedStringTypeDeserialization()
                 .Build();
 
-            var expected = yaml;
-
-            var result = deserializer.Deserialize<object>(new StringReader(expected));
+            var result = deserializer.Deserialize<object>(new StringReader(yaml));
 
             result.Should().Be(value);
         }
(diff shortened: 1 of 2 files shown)


EdwardCooke:
I don't think an unisigned long is the most widely used number type. I would think that int is the most widely used as that is what the compiler seems to default as, example: `var n = 0`, `n` is an `int`. This would also end up being a breaking change due to changing the return types of the method being different. If you change this to doing an int, then long, then unsigned long I'd be happy to merge this in with the next breaking change build, but, I'm hoping that it will be a while before that happens again. The other alternative would be to change this up to be an option that can be set on the deserializerbuilder, something like a `SmallestNumericDataType`. Probably an option on the same method that enables the type conversion. Can't remember the name right off the top.

PR #954, opened 2024-08-15
Avoid closures in CachedTypeInspector

MattKotsenas:
The `CachedTypeInspector` creates a closure in order to capture the method parameters. This results in 2 allocations per object serialized. I switched to using the `ConcurrentDictionary.GetOrAdd()` method that takes a third context parameter in order to avoid allocations.

I also added an extension method as described in https://github.com/dotnet/runtime/issues/13978 for runtimes where the overload doesn't exist. I also updated the language version from 8 to 9 in order to get access to [static lambdas](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/static-anonymous-functions) in order to ensure that the lambdas don't capture.

This reduces allocations by about ~4% in the benchmark:

## Before

```
// * Summary *

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22635.4010)
Intel Core i9-10940X CPU 3.30GHz, 1 CPU, 28 logical and 14 physical cores
.NET SDK 9.0.100-preview.7.24407.12
  [Host]                       : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET 8.0           : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET Framework 4.7 : .NET Framework 4.8.1 (4.8.9261.0), X64 RyuJIT VectorSize=256

IterationCount=15  LaunchCount=2  WarmupCount=10

| Method     | Job                          | Runtime            | Mean      | Error    | StdDev   | Gen0      | Gen1     | Gen2     | Allocated |
|----------- |----------------------------- |------------------- |----------:|---------:|---------:|----------:|---------:|---------:|----------:|
| Serializer | MediumRun-.NET 8.0           | .NET 8.0           |  50.25 ms | 1.055 ms | 1.578 ms | 2000.0000 | 500.0000 |        - |  24.44 MB |
| Serializer | MediumRun-.NET Framework 4.7 | .NET Framework 4.7 | 118.87 ms | 3.662 ms | 5.367 ms | 7800.0000 | 600.0000 | 200.0000 |     48 MB |

// * Warnings *
MinIterationTime
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0 -> The minimum observed iteration time is 93.442ms which is very small. It's recommended to increase it to at least 100ms using more operations.
```

## After

```
[...]

PR #953, opened 2024-08-15
Use Attribute.GetCustomAttributes to reduce allocations / improve performance

MattKotsenas:
Because the project no longer targets `netstandard1.3`, we can replace the custom implementation inside `ReflectionExtensions.GetAllCustomAttributes()` with `Attribute.GetCustomAttributes()`. 

In addition to simpler code, in the provided benchmark this eliminates 4 `List<Attribute>` allocations _per serialized element_. In the benchmark that results in ~10% less CPU time and ~5% fewer allocations.

Here's the added BenchmarkDotNet results on my machine:

### Before

```
// * Summary *

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22635.4010)
Intel Core i9-10940X CPU 3.30GHz, 1 CPU, 28 logical and 14 physical cores
.NET SDK 9.0.100-preview.7.24407.12
  [Host]                       : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET 8.0           : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET Framework 4.7 : .NET Framework 4.8.1 (4.8.9261.0), X64 RyuJIT VectorSize=256

IterationCount=15  LaunchCount=2  WarmupCount=10

| Method     | Job                          | Runtime            | Mean      | Error    | StdDev   | Gen0      | Gen1     | Gen2     | Allocated |
|----------- |----------------------------- |------------------- |----------:|---------:|---------:|----------:|---------:|---------:|----------:|
| Serializer | MediumRun-.NET 8.0           | .NET 8.0           |  48.12 ms | 1.434 ms | 2.057 ms | 2000.0000 | 500.0000 |        - |  24.44 MB |
| Serializer | MediumRun-.NET Framework 4.7 | .NET Framework 4.7 | 108.69 ms | 2.218 ms | 3.181 ms | 7800.0000 | 600.0000 | 200.0000 |     48 MB |

// * Warnings *
MinIterationTime
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0 -> The minimum observed iteration time is 88.232ms which is very small. It's recommended to increase it to at least 100ms using more operations.

// * Hints *
Outliers
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0 -> 2 outliers were removed (55.60 ms, 58.61 ms)
```

### After

```
[...]

PR #956, opened 2024-08-16
Cache lookups to IEnumerable<IYamlTypeConverter>

MattKotsenas:
Several types accept an `IEnumerable<IYamlTypeConverter>`, and then repeatedly traverse the list searching for a converter that accepts the type to be serialized. Traversing an IEnumerable with LINQ results in an enumerator (to walk the `IEnumerable`) and a Func (to filter the enumerable) allocation per visitor per object serialized. In the benchmark setup this results in 12 allocations per object serialized.

To avoid the repeated lookups and allocations, I introduced an internal `TypeConverterCache` that accepts the collection of type converters and maps the type to convert to the first `IYamlTypeConverter` that accepts the type (the same as the previous behavior). This eliminates all 12 allocations and has a small CPU time win as well.

To avoid a breaking change, I left the `protected IEnumerable<IYamlTypeConverter>` and added a `private TypeConverterCache` on `PreProcessingPhaseObjectGraphVisitorSkeleton`. If / when we want to remove cruft and can accept breaking changes we can remove this field.

Note that each type that previously was doing list traversal has its own cache. If it's OK to change the constructor signature on these types, we could instead construct a single cache in the builder (e.g. in `BuilderSkeleton.BuildTypeConverters`) and pass it in. However, I assume the builder is designed for public extension, so I avoided that in this change.

In the provided benchmark, this results in a 15% - 25% reduction in allocations:

## Before

```
// * Summary *

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22635.4010)
Intel Core i9-10940X CPU 3.30GHz, 1 CPU, 28 logical and 14 physical cores
.NET SDK 9.0.100-preview.7.24407.12
  [Host]                       : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET 8.0           : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET Framework 4.7 : .NET Framework 4.8.1 (4.8.9261.0), X64 RyuJIT VectorSize=256

IterationCount=15  LaunchCount=2  WarmupCount=10

| Method     | Job                          | Runtime            | Mean      | Error    | StdDev   | Gen0      | Gen1     | Gen2     | Allocated |
|----------- |----------------------------- |------------------- |----------:|---------:|---------:|----------:|---------:|---------:|----------:|
| Serializer | MediumRun-.NET 8.0           | .NET 8.0           |  49.16 ms | 1.436 ms | 2.105 ms | 2000.0000 | 500.0000 |        - |  24.44 MB |
| Serializer | MediumRun-.NET Framework 4.7 | .NET Framework 4.7 | 109.79 ms | 2.546 ms | 3.569 ms | 7800.0000 | 600.0000 | 200.0000 |     48 MB |

// * Warnings *
MinIterationTime
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0 -> The minimum observed iteration time is 91.906ms which is very small. It's recommended to increase it to at least 100ms using more operations.

// * Hints *
Outliers
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0           -> 1 outlier  was  removed (57.76 ms)
  SerializationBenchmarks.Serializer: MediumRun-.NET Framework 4.7 -> 2 outliers were removed (121.75 ms, 123.43 ms)
```

## After

```
// * Summary *

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22635.4010)
Intel Core i9-10940X CPU 3.30GHz, 1 CPU, 28 logical and 14 physical cores
.NET SDK 9.0.100-preview.7.24407.12
  [Host]                       : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET 8.0           : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  MediumRun-.NET Framework 4.7 : .NET Framework 4.8.1 (4.8.9261.0), X64 RyuJIT VectorSize=256

IterationCount=15  LaunchCount=2  WarmupCount=10

| Method     | Job                          | Runtime            | Mean      | Error    | StdDev   | Gen0      | Gen1     | Gen2     | Allocated |
|----------- |----------------------------- |------------------- |----------:|---------:|---------:|----------:|---------:|---------:|----------:|
| Serializer | MediumRun-.NET 8.0           | .NET 8.0           |  46.67 ms | 1.620 ms | 2.425 ms | 1500.0000 | 500.0000 |        - |  17.95 MB |
| Serializer | MediumRun-.NET Framework 4.7 | .NET Framework 4.7 | 108.68 ms | 3.342 ms | 5.002 ms | 6600.0000 | 600.0000 | 200.0000 |  41.49 MB |

// * Warnings *
MultimodalDistribution
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0 -> It seems that the distribution can have several modes (mValue = 2.83)
MinIterationTime
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0 -> The minimum observed iteration time is 85.323ms which is very small. It's recommended to increase it to at least 100ms using more operations.

// * Hints *
Outliers
  SerializationBenchmarks.Serializer: MediumRun-.NET 8.0           -> 1 outlier  was  removed (52.89 ms)
  SerializationBenchmarks.Serializer: MediumRun-.NET Framework 4.7 -> 1 outlier  was  removed (122.13 ms)
```

### DIFF
--- a/YamlDotNet/Serialization/Utilities/TypeConverterCache.cs
+++ b/YamlDotNet/Serialization/Utilities/TypeConverterCache.cs
@@ -1,0 +1,104 @@
+﻿// This file is part of YamlDotNet - A .NET library for YAML.
+// Copyright (c) Antoine Aubry and contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+// of the Software, and to permit persons to whom the Software is furnished to do
+// so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+
+namespace YamlDotNet.Serialization.Utilities
+{
+    /// <summary>
+    /// A cache / map for <see cref="IYamlTypeConverter"/> instances.
+    /// </summary>
+    internal sealed class TypeConverterCache
+    {
+        private readonly IYamlTypeConverter[] typeConverters;
+        private readonly Dictionary<Type, (bool HasMatch, IYamlTypeConverter TypeConverter)> cache = new();
+
+        public TypeConverterCache(IEnumerable<IYamlTypeConverter>? typeConverters)
+        {
+            this.typeConverters = typeConverters != null
+                ? typeConverters.ToArray()
+                : Array.Empty<IYamlTypeConverter>();
+        }
+
+        /// <summary>
+        /// Returns the first <see cref="IYamlTypeConverter"/> that accepts the given type.
+        /// </summary>
+        /// <param name="type">The <see cref="Type"/> to lookup.</param>
+        /// <param name="typeConverter">The <see cref="IYamlTypeConverter" /> that accepts this type or <see langword="false" /> if no converter is found.</param>
+        /// <returns><see langword="true"/> if a type converter was found; <see langword="false"/> otherwise.</returns>
+        public bool TryGetConverterForType(Type type, [NotNullWhen(true)] out IYamlTypeConverter? typeConverter)
+        {
+            if (cache.TryGetValue(type, out var result))
+            {
+                typeConverter = result.TypeConverter;
+                return result.HasMatch;
+            }
+
+            typeConverter = LookupTypeConverter(type);
+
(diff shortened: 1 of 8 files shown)


MattKotsenas:
Note that this PR uses the same benchmark introduced in #953 and #954. So if / when one of these PRs merge, the other should be rebased on top of latest main to remove the duplicate commit.

I'm happy to do this housekeeping. I just mention it for clarity, and so that each PR can be reviewed independently, rather than forcing them to all be chained together.

MattKotsenas:
Alternatively, there aren't many `IYamlTypeConverter`s built-in, and the `Accept()` call is cheap, thus we can get similar benefit by changing the `IEnumerable` to be an array instead and manually unrolling the LINQ calls.

I didn't go that route initially because:

1. It's still a breaking change
2. It's less DRY, as each consumer (and any new consumers in the future) would need to also know to avoid LINQ

If you have a strong preference of either path let me know and I'm happy to make updates here.

EdwardCooke:
I'm happier with no breaking changes, even in the constructors, but yes, this is public so it can be extended. In fact I saw a consumer of yamldotnet the other day that inherited some of these low level classes and made minor changes. Like the kubernetes client.

YamlDotNet/Serialization/Utilities/TypeConverterCache.cs
```diff
@@ -0,0 +1,106 @@
...
+
+namespace YamlDotNet.Serialization.Utilities
+{
+    /// <summary>
+    /// A cache / map for <see cref="IYamlTypeConverter"/> instances.
+    /// </summary>
+    internal sealed class TypeConverterCache
+    {
+        private readonly IYamlTypeConverter[] typeConverters;
+        private readonly Dictionary<Type, (bool HasMatch, IYamlTypeConverter TypeConverter)> cache = new();
```
EdwardCooke, line 35: I'm going to make a PR to fix this, but this should have been a concurrent dictionary. It's failing on my laptop because it's fast enough to catch it where the build system is slow enought that it didn't. Found it while testing your other PR for the build properties.

MattKotsenas: You're correct, sorry that I missed this.

PR #973, opened 2024-09-06
Convert to use Central Package Management

lahma:
Now single place to have packages and their versions.  See https://devblogs.microsoft.com/nuget/introducing-central-package-management/ for information about the feature.


EdwardCooke:
This is an interesting idea. In the most recently merged PR we cleaned up some unused references, can you rebase?

lahma:
Rebased, I removed some unnecessary references from projects and also targeted benchmark projects just against net8.0 for easiest dependency resolving. I think it should suffice to benchmark against modern runtime.

EdwardCooke:
Thanks. However. We do benchmark against net47. I do agree targeting the latest net 80 framework as well.

lahma:
I restored the `net47` target (I think this should ideally be `net472` as 4.7.2 is mainstream and comes with the OSs).

PR #990, opened 2024-10-09
perf: Optimize regular numbers parse logics

filzrev:
This PR intended to optimize performance for number deserialization by changing Parse to `TryParse` method. 
It's discussed at https://github.com/aaubry/YamlDotNet/pull/794

Additionally this PR fix `double` value deserialization problems that are found on UnitTests.

**What's Changes**
- Replace existing regular number `Parse` logics with `TryParse` method.
- Separate `float`/`double` parse logics for `.NET Framework`.
- Add code path for .NET that parse text as `double`, then checks float value ranges. 
- Add related unit tests (`UnquotedStringTypeDeserialization_RegularNumbers`)

**.NET Behavior differences**
On .NET Framework.  `float.Parse((double.MaxValue + 1).ToString())` throw `OverflowException`.
But on .NET Core. It returns PosititiveInfinite(`∞`).

These behavior differences are described at [Single.Parse Method](https://learn.microsoft.com/en-us/dotnet/api/system.single.parse?view=net-8.0) page.

> In .NET Core 3.0 and later, values that are too large to represent are rounded to [PositiveInfinity] or
> [NegativeInfinity](https://learn.microsoft.com/en-us/dotnet/api/system.single.negativeinfinity?view=net-8.0#system-single-negativeinfinity) as required by the IEEE 754 specification. 
> In prior versions, including .NET Framework, parsing a value that was too large to represent resulted in failure.

On current implementation. 
Values outside the range of the `float` type are converted to `-∞`/`∞`.
So I've added logics to check float value ranges.

### DIFF
--- a/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
+++ b/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
@@ -382,17 +382,16 @@ namespace YamlDotNet.Serialization.NodeDeserializers
                     }
                     else if (Regex.IsMatch(v, @"[-+]?(\.[0-9]+|[0-9]+(\.[0-9]*)?)([eE][-+]?[0-9]+)?")) //regular number
                     {
-#pragma warning disable format
-                        if      ( byte.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var byteValue  )) { result = byteValue;  }
-                        else if (short.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var shortValue )) { result = shortValue; }
-                        else if (  int.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var intValue   )) { result = intValue;   }
-                        else if ( long.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var longValue  )) { result = longValue;  }
-                        else if (ulong.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var ulongValue )) { result = ulongValue; }
+                        if (byte.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var byteValue)) { result = byteValue; }
+                        else if (short.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var shortValue)) { result = shortValue; }
+                        else if (int.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var intValue)) { result = intValue; }
+                        else if (long.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var longValue)) { result = longValue; }
+                        else if (ulong.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var ulongValue)) { result = ulongValue; }
 #if NETFRAMEWORK
-                        else if ( float.TryParse(v, NumberStyles.Float | NumberStyles.AllowThousands, formatter.NumberFormat, out var floatValue)) { result = floatValue; }
-                        else if (double.TryParse(v, NumberStyles.Float | NumberStyles.AllowThousands, formatter.NumberFormat, out var doubleValue)) { result = doubleValue; }
+                        else if (float.TryParse(v, NumberStyles.Float, formatter.NumberFormat, out var floatValue)) { result = floatValue; }
+                        else if (double.TryParse(v, NumberStyles.Float, formatter.NumberFormat, out var doubleValue)) { result = doubleValue; }
 #else
-                        else if (double.TryParse(v, NumberStyles.Float | NumberStyles.AllowThousands, formatter.NumberFormat, out var doubleValue))
+                        else if (double.TryParse(v, NumberStyles.Float, formatter.NumberFormat, out var doubleValue))
                         {
                             var floatValue = (float)doubleValue;
                             if (!float.IsNaN(floatValue) && !float.IsInfinity(floatValue)) // .NET 6 or later support float.IsNormal
@@ -405,7 +404,6 @@ namespace YamlDotNet.Serialization.NodeDeserializers
                             }
                         }
 #endif
-#pragma warning restore format
                         else
                         {
                             //we couldn't parse it, default to string, It's probably too big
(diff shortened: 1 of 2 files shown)


filzrev:
### Microbenchmark results

`Optimized` version run about 10x faster. and memory allocation is reduced about 50%.

**Baseline**

| Method      | Job                         | Runtime            | Mean     | Error    | StdDev  | Exceptions | Gen0   | Gen1   | Allocated |
|------------ |---------------------------- |------------------- |---------:|---------:|--------:|-----------:|-------:|-------:|----------:|
| Deserialize | ShortRun-.NET 8.0           | .NET 8.0           | 176.9 us | 18.02 us | 0.99 us |    18.0000 | 7.3242 |      - |  29.96 KB |
| Deserialize | ShortRun-.NET Framework 4.7 | .NET Framework 4.7 | 588.5 us | 60.13 us | 3.30 us |    23.0000 | 9.7656 | 0.9766 |  43.63 KB |

**Optimized**
| Method      | Job                         | Runtime            | Mean     | Error    | StdDev   | Exceptions | Gen0   | Gen1   | Allocated |
|------------ |---------------------------- |------------------- |---------:|---------:|---------:|-----------:|-------:|-------:|----------:|
| Deserialize | ShortRun-.NET 8.0           | .NET 8.0           | 20.88 us | 3.234 us | 0.177 us |          - | 4.0283 |      - |  16.45 KB |
| Deserialize | ShortRun-.NET Framework 4.7 | .NET Framework 4.7 | 52.78 us | 9.504 us | 0.521 us |          - | 5.3711 | 0.1221 |  23.66 KB |

**Test Code**
```csharp
[MemoryDiagnoser]
[ExceptionDiagnoser]
[ShortRunJob(RuntimeMoniker.Net80)]
[ShortRunJob(RuntimeMoniker.Net47)]
public class AttemptUnknownTypeDeserializationBenchmarks
{
    private static readonly IDeserializer deserializer = new DeserializerBuilder().WithAttemptingUnquotedStringTypeDeserialization().Build();

    private static readonly string TestData =
        """
        byteMaxValue:   255
        shortMaxValue:  32767
        longMaxValue :  9223372036854775807
        ulongMaxValue:  18446744073709551615
        floatMaxValue:  3.4028235E+38
        doubleMaxValue: 1.7976931348623157E+308
        """;

    [Benchmark]
    public object? Deserialize()
    {
        return deserializer.Deserialize(TestData);
    }
}

```

YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
```diff
@@ -382,13 +382,30 @@ private static object CastInteger(ulong number, TypeCode typeCode)
...
                     else if (Regex.IsMatch(v, @"[-+]?(\.[0-9]+|[0-9]+(\.[0-9]*)?)([eE][-+]?[0-9]+)?")) //regular number
                     {
-                        if (TryAndSwallow(() => byte.Parse(v, formatter.NumberFormat), out result)) { }
-                        else if (TryAndSwallow(() => short.Parse(v, formatter.NumberFormat), out result)) { }
-                        else if (TryAndSwallow(() => int.Parse(v, formatter.NumberFormat), out result)) { }
-                        else if (TryAndSwallow(() => long.Parse(v, formatter.NumberFormat), out result)) { }
-                        else if (TryAndSwallow(() => ulong.Parse(v, formatter.NumberFormat), out result)) { }
-                        else if (TryAndSwallow(() => float.Parse(v, formatter.NumberFormat), out result)) { }
-                        else if (TryAndSwallow(() => double.Parse(v, formatter.NumberFormat), out result)) { }
+#pragma warning disable format
```
EdwardCooke, line 385: We need to keep the formatting in place please.

YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
```diff
@@ -382,13 +382,30 @@ private static object CastInteger(ulong number, TypeCode typeCode)
...
+                        if      ( byte.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var byteValue  )) { result = byteValue;  }
+                        else if (short.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var shortValue )) { result = shortValue; }
+                        else if (  int.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var intValue   )) { result = intValue;   }
+                        else if ( long.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var longValue  )) { result = longValue;  }
+                        else if (ulong.TryParse(v, NumberStyles.Integer, formatter.NumberFormat, out var ulongValue )) { result = ulongValue; }
+#if NETFRAMEWORK
+                        else if ( float.TryParse(v, NumberStyles.Float | NumberStyles.AllowThousands, formatter.NumberFormat, out var floatValue)) { result = floatValue; }
```
EdwardCooke, line 392: Is there a reason for not using `formatter.NumberFormat` here and below?

EdwardCooke, line 392: According to the spec, float data types shouldn't allow commas, I'm assuming that's what the `AllowThousands` does.

https://yaml.org/type/float.html

EdwardCooke:
I like performance fixes. And this is a pretty big one! If you can address the 2 comments then I'll merge this in and get it out.

YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
filzrev: I've removed `pragma warning disable format` and format code with default settings

YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs
filzrev: I've removed `NumberStyles.AllowThousands` flags.

filzrev:
Apart from this PR.
I've noticed another performance relating issue.

https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Serialization/NodeDeserializers/ScalarNodeDeserializer.cs#L383

Above regex checking `regular number`(integer or floating number).
But `any text that contains digits` match regex.
So unneeded `Parse` operations are called multiple times.

Is it able to change Regex to use more strict regex with `^` and `$` ?
or is there any YAML specs that require text partial match?

EdwardCooke:
You can add those to that regex without any issue I would think.

filzrev:
> You can add those to that regex without any issue I would think.

Thanks for your response.
I'll try to create another PR after this PR is merged.

filzrev:
I've removed wrong comment in commit.

`float.IsNormal` is available .NET Core 2.1 and The behavior will be different from what I'm expecting.
https://learn.microsoft.com/en-us/dotnet/api/system.single.isnormal?view=net-8.0#remarks

PR #1000, opened 2024-10-25
Bring back stack trace to YamlException.ToString

### LINKED ISSUE #999: YamlException.ToString hiddes stack track
The `YamlException` class overrides the `ToString` method so that it only displays parsing error of the exception. Typically, in such cases, exceptions also include the StackTrace by default. This change makes debugging more challenging, especially if the developer is unaware that the ToString method’s behavior is non-standard.

rosieks:
This PR introduce `Reason` property in `YamlException` to bring back default functionality of `ToString` method.
Fixes #999

### DIFF
--- a/YamlDotNet.Test/Core/YamlExceptionTests.cs
+++ b/YamlDotNet.Test/Core/YamlExceptionTests.cs
@@ -29,35 +29,35 @@ namespace YamlDotNet.Test.Core
     public class YamlExceptionTests
     {
         [Fact]
-        public void VerifyToStringWithEmptyMarks()
+        public void VerifyMessageWithEmptyMarks()
         {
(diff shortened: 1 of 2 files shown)


rosieks:
Testing `ToString` doesn't make sense. It's inherited behavior and it's tested already somewhere else.

liskin:
> Testing `ToString` doesn't make sense. It's inherited behavior and it's tested already somewhere else.

Suppose that's right. I saw it as a way to express "desired behavior". Sure, it's obvious that the code does what it should, it's a typed language, but having a test case saying "YamlException stringification doesn't hide the inner exception" seemed as a useful regression test. Doesn't matter that much though.

TheBigNeo (approved):

PR #1007, opened 2024-11-10
perf: Optimize number parse logics

filzrev:
This PR intended to fix following issues.

**1. Change number checks to more strict regex**
It's discussed at https://github.com/aaubry/YamlDotNet/pull/990#issuecomment-2405797524
When parsing scalar values. whitespaces are already ignored by scanner.
So there is no needs using  `partial regex match`.

**2. Change Hex string format parse logics not to throw exceptions**
Same as integer/float string format. (That is fixed at #990)

### DIFF
--- a/YamlDotNet.Test/Serialization/DeserializerTest.cs
+++ b/YamlDotNet.Test/Serialization/DeserializerTest.cs
@@ -259,6 +259,27 @@ b: &number 1
         }
 
         [Theory]
+        [InlineData(System.Byte.MinValue)]
+        [InlineData(System.Byte.MaxValue)]
+        [InlineData(System.Int16.MinValue)]
+        [InlineData(System.Int16.MaxValue)]
+        [InlineData(System.Int32.MinValue)]
+        [InlineData(System.Int32.MaxValue)]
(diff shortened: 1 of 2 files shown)


YamlDotNet.Test/Serialization/DeserializerTest.cs
```diff
@@ -258,6 +258,27 @@ public void UnquotedStringTypeDeserialization_RegularNumbers(object expected)
...
 
+        [Theory]
+        [InlineData(System.Byte.MinValue)]
+        [InlineData(System.Byte.MaxValue)]
+        [InlineData(System.Int16.MinValue)]
+        [InlineData(System.Int16.MaxValue)]
+        [InlineData(System.Int32.MinValue)]
+        [InlineData(System.Int32.MaxValue)]
+        [InlineData(System.Int64.MinValue)]
+        [InlineData(System.Int64.MaxValue)]
```
filzrev, line 269: UnitTest for `ulong` is not included.
Because `ulong` code path is not seems not be used. 
Hex/Octal number representation is same for `long`/`ulong` and these number is always parsed as `long`

PR #1032, opened 2025-01-16
fix #1031

### LINKED ISSUE #1031: [AoT] Compile error if only one StaticContext exist and in global namespace
**Describe the bug**
In the project, if there is a unique class with `YamlStaticContext` attribute, and this class is in the global namespace, while there are other classes with `YamlSerializable` attribute, the compilation will fail.

**To Reproduce**
1. Create a C# console project and enable NativeAoT publish. ( __Dont__ trigger `Do not use top-level statements`)
2. Add necessary package from nuget. (`YamlDotNet` and `Vecc.YamlDotNet.Analyzers.StaticGenerator`)
3. Delete all template content in `Program.cs`
4. Paste code.
```CSharp
Console.WriteLine("Hello world");

[YamlDotNet.Serialization.YamlSerializable]
public class Model {}

[YamlDotNet.Serialization.YamlStaticContext]
public partial class YamlContext : YamlDotNet.Serialization.StaticContext {}
```

Build output here
```
$ dotnet build
Restore complete (2.2s)
  BugReproduceProject failed with 1 error(s) (0.4s)
C:\BugReproduceProject\obj\Debug\net9.0\YamlDotNet.Analyzers.StaticGenerator\YamlDotNet.Analyzers.StaticGenerator.TypeFactoryGenerator\YamlDotNetAutoGraph.g.cs(8,10): error CS1001: Identifier expected

Build failed with 1 error(s) in 2.9s
```

dogdie233:
## What cause the issue

In line 94

https://github.com/aaubry/YamlDotNet/blob/b8ac2a98ffcc12434eff6c6abb75b38ad1b1ab04/YamlDotNet.Analyzers.StaticGenerator/TypeFactoryGenerator.cs#L92-L98

When the StaticContext to be generated is in the global namespace, `syntaxReceiver.YamlStaticContextType?.GetNamespace()` returns an empty string instead of null. As a result, the `??` operator does not take effect.

So, the generated code in `YamlDotNetAutoGraph.g.cs` ends up as follows:
```CSharp
...
namespace 
{
    ...
```
Since there is no identifier following `namespace`, the compilation fails.

## About unit test case

Due to the stringent conditions required to trigger this issue, I am unsure how to write a unit test. Should I create a new console project to reproduce the issue?

## A potential problem

When the auto-generated `StaticContext` is in the global namespace, other classes, such as `StaticObjectFactory`, will also be created in the global namespace. This could potentially conflict with other classes.

### DIFF
--- a/YamlDotNet.Analyzers.StaticGenerator/TypeFactoryGenerator.cs
+++ b/YamlDotNet.Analyzers.StaticGenerator/TypeFactoryGenerator.cs
@@ -89,10 +89,14 @@ namespace YamlDotNet.Analyzers.StaticGenerator
                 write("using System;", true);
                 write("using System.Collections.Generic;", true);
 
-                var namespaceName = syntaxReceiver.YamlStaticContextType?.ContainingNamespace.ContainingNamespace;
+                var namespaceName = syntaxReceiver.YamlStaticContextType?.ContainingNamespace?.Name;
+                var isGlobalNamespace = string.IsNullOrEmpty(namespaceName);
 
-                write($"namespace {syntaxReceiver.YamlStaticContextType?.GetNamespace() ?? "YamlDotNet.Static"}", true);
-                write("{", true); indent();
+                if (!isGlobalNamespace)
+                {
+                    write($"namespace {namespaceName}", true);
+                    write("{", true); indent();
+                }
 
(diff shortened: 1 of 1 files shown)


dogdie233:
oops. im sorry forgot to checkout a new branch

Issue #1045, opened 2025-02-27
How do I handle the content represented by an anchor in an IYamlConverter?


Polar-Pumpkin:
It seems that the anchor of yaml cannot be used to insert content into a list, so I want to reference repeated content and add new content to the list by converting it into a nested list.

For example, considering this yaml:
```yaml
executors:
  - name: a
    filters: &shared
      - type: foo
        value: 1
      - type: bar
        value: 2
  - name: b
    filters: 
      - *shared
      - type: extra
        value: 3
```

I expected to parsing like:
```yaml
executors:
  - name: a
    filters: &shared
      - type: foo
        value: 1
      - type: bar
        value: 2
  - name: b
    filters: 
      - - type: foo
          value: 1
        - type: bar
          value: 2
      - type: extra
        value: 3
````

I wrote an IYamlConverter to manually parse this list. Originally, I thought that the anchor would be directly converted into a repeated `ParsingEvent` queue, but what I actually got was an `AnchorAlias` ​​event and the name of the anchor. How can I get the content represented by this anchor?

This code will getting exception: `Unhandled exception. (Line: 1, Col: 1, Idx: 0) - (Line: 1, Col: 1, Idx: 0): No node deserializer was able to deserialize the node into type YamlAnchorTest.Filter, YamlAnchorTest, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null`

```csharp
using YamlDotNet.Core;
using YamlDotNet.Core.Events;
using YamlDotNet.Serialization;
using YamlDotNet.Serialization.NamingConventions;

namespace YamlAnchorTest;

public class Configuration {
    public List<Executor> Executors { get; set; }
}

public class Executor {
    public string Name { get; set; }
    public List<Filter> Filters { get; set; }
}

public class Filter {
    public string Type { get; set; }
    public int Value { get; set; }
}

public class FilterCollectionConverter : IYamlTypeConverter {
    public static IYamlTypeConverter Instance { get; } = new FilterCollectionConverter();

    public bool Accepts(Type type) {
        return type == typeof(List<Filter>);
    }

    public object ReadYaml(IParser parser, Type type, ObjectDeserializer rootDeserializer) {
        var result = new List<Filter>();
        parser.Consume<SequenceStart>();
        while (!parser.TryConsume<SequenceEnd>(out _)) {
            // Got AnchorAlias here! With nothing else but anchor name?
            if (parser.TryConsume<SequenceStart>(out _)) {
[...]

Issue #1043, opened 2025-02-27
Duplicate contents represented by Anchor are actually the same reference object


Polar-Pumpkin:
**Describe the bug**
I tried to use anchor to simplify the repetitive content in my yaml configuration. I thought anchor just represented repeated yaml text content, but I found that the objects represented by anchor all refer to the same object, rather than repeatedly creating different (reference) objects as I imagined.

If the deserialized data is read-only, there should be no problem, but occasionally I want to store some states in the deserialized object. At this time, some conflicts will occur due to they are actually the same reference object.

I want to know if this behavior is designed? Is my understanding of anchor wrong?
Thank you very much for your answer and help. ❤ 

**To Reproduce**
```csharp
using YamlDotNet.Serialization;
using YamlDotNet.Serialization.NamingConventions;

namespace YamlAnchorTest;

public class Configuration {
    public List<Executor> Executors { get; set; }
}

public class Executor {
    public string Name { get; set; }
    public List<Filter> Filters { get; set; }
}

public class Filter {
    public string Type { get; set; }
    public int Value { get; set; }
}

internal static class Program {
    private static void Main(string[] args) {
        var document =
            """
            executors:
              - name: a
                filters: &shared
                  - type: foo
                    value: 1
                  - type: bar
                    value: 2
              - name: b
                filters: *shared
            """;
        var deserializer = new DeserializerBuilder()
            .WithNamingConvention(HyphenatedNamingConvention.Instance)
            .WithEnumNamingConvention(HyphenatedNamingConvention.Instance)
            .Build();
        var conf = deserializer.Deserialize<Configuration>(document);
        Console.WriteLine(ReferenceEquals(conf.Executors[0].Filters, conf.Executors[1].Filters));
        // Output: true
        // If I add an item to the Filters of one Executor at this time, the other Executor will also changed
    }
}
```

PR #1046, opened 2025-02-28
Replay buffered parser to create different (reference) objects at alias

Polar-Pumpkin:
In response to the issues mentioned in #1043 and #1045, I have written two corresponding unit tests. The solution is to cache the buffered parsers in a way similar to `ITypeDiscriminatingNodeDeserializer`, and then replay the buffer when unfolding aliases to recreate objects with different reference addresses.

While running the unit tests, I noticed that YAML allows self-referencing anchors to create cycles. Therefore, I added a check that uses the original cached-object-based solution when a self-reference is detected.

After the fix, the unit test `RoundtripArrayOfIdenticalObjects` failed because the objects were serialized as anchors, representing the same references. However, during deserialization, different references were generated (although the content was the same), causing the test to fail. This left me unsure whether what I did was correct: do anchors only represent objects with the same reference?

That being said, I just wanted to use anchors to simplify the parts of my configuration file that require duplication. If you're open to my understanding of anchors, I could also implement this as an switch in builder.

Thank you very much for any support and help you provide! ❤️


EdwardCooke:
I'm open to your understanding, my question is that since anchors and aliases are effectively pointing to the same "object" in the yaml it makes since that they are the same reference object. If you made this an opt-in option then I'd be much more keen on merging this in. Until that happens though, I'm not going to.

Polar-Pumpkin:
Thank you very much <3 I have submitted the relevant changes to make it an optional feature.

PR #1054, opened 2025-05-13
Use string interning for fun and performance

simonthum:
This cuts about 20% of parsing cost and probably more in final object graph size, depending on input data.

In terms of the LoadLarge benchmark:

| Method           | Mean     | Error   | StdDev   | Gen0      | Gen1      | Gen2      | Allocated |
|-----------------:|---------:|--------:|---------:|----------:|----------:|----------:|----------:|
| LoadLarge before | 136.9 ms | 3.11 ms | 9.03 ms  | 7500.0000 | 3500.0000 | 1500.0000 |  39.65 MB |
| LoadLarge after  | 114.4 ms | 3.70 ms | 10.67 ms | 6000.0000 | 2000.0000 | 1000.0000 |  39.65 MB |

This is just a simple string.Intern call. It might make sense to employ or pass a pool, like a XmlNameTable does with XML, to save on scalar objects not just strings.

As for my motivation, yaml parsing dominates a startup time I care about. This seems like an easy fix.

### DIFF
--- a/YamlDotNet/Core/AnchorName.cs
+++ b/YamlDotNet/Core/AnchorName.cs
@@ -39,12 +39,17 @@ namespace YamlDotNet.Core
 
         public AnchorName(string value)
         {
-            this.value = value ?? throw new ArgumentNullException(nameof(value));
+            if (value != null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
 
             if (!AnchorPattern.IsMatch(value))
             {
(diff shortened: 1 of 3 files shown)


simonthum:
The build failure seems unrelated to the changes.

EdwardCooke:
It's because your branch name is master. If you use a branch not named master we can work on getting this merged in.

simonthum:
I opened another PR to change the branch.

PR #1055, opened 2025-05-14
Use string interning

simonthum:
This cuts about 20% of parsing cost and probably more in final object graph size, depending on input data.

In terms of the LoadLarge benchmark:

| Method           | Mean     | Error   | StdDev   | Gen0      | Gen1      | Gen2      | Allocated |
|-----------------:|---------:|--------:|---------:|----------:|----------:|----------:|----------:|
| LoadLarge before | 136.9 ms | 3.11 ms | 9.03 ms  | 7500.0000 | 3500.0000 | 1500.0000 |  39.65 MB |
| LoadLarge after  | 114.4 ms | 3.70 ms | 10.67 ms | 6000.0000 | 2000.0000 | 1000.0000 |  39.65 MB |

This is just a simple string.Intern call. It might make sense to emplyor pass a pool, like a XmlNameTable does with XML, to save on scalar objects not just strings.

### DIFF
--- a/YamlDotNet/Core/Events/Scalar.cs
+++ b/YamlDotNet/Core/Events/Scalar.cs
@@ -79,7 +79,7 @@ namespace YamlDotNet.Core.Events
         public Scalar(AnchorName anchor, TagName tag, string value, ScalarStyle style, bool isPlainImplicit, bool isQuotedImplicit, Mark start, Mark end, bool isKey = false)
             : base(anchor, tag, start, end)
         {
-            this.Value = value;
+            this.Value = (isKey || value.Length < 40) ? string.Intern(value) : value;
             this.Style = style;
             this.IsPlainImplicit = isPlainImplicit;
             this.IsQuotedImplicit = isQuotedImplicit;
--- a/YamlDotNet/Core/TagName.cs
+++ b/YamlDotNet/Core/TagName.cs
@@ -39,7 +39,7 @@ namespace YamlDotNet.Core
 
         public TagName(string value)
         {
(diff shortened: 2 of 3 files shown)


simonthum:
I'm not sure what formatting I broke. I tried to improve, but l 51 is just a brace.

YamlDotNet/Core/Events/Scalar.cs
EdwardCooke, line 1: What’s the purpose of checking if it’s 40 or less?

YamlDotNet/Core/TagName.cs
EdwardCooke, line 1: We should still throw the argumentnullexception if it’s null here

EdwardCooke (changes requested):

YamlDotNet/Core/TagName.cs
EdwardCooke: And the IsKey?

simonthum: Intern() does that, but it should not hurt to have it explicit.

YamlDotNet/Core/Events/Scalar.cs
EdwardCooke: Let’s try it without that. I know the files I typically deal with they do not repeat.

YamlDotNet/Core/TagName.cs
EdwardCooke: Let’s do explicit so we know exactly what the exception will be.

YamlDotNet/Core/Events/Scalar.cs
simonthum: Done

PR #1064, opened 2025-08-26
Nullable fixes in non-public code

Kielek:
This PR removes a lot of warnings related to nullable annotations. All of them are related only to non-public contracts.

It can be fixed also for public classes, but it can be considered as a braking changes, so it should be candidate for separate PR.

While reviewing, you can consider commit by commit reviews.


Kielek:
@EdwardCooke?

PR #1072, opened 2026-01-05
Allow specifying a maximum recursion for the deserializer

aaubry:
This adds a `WithMaximumRecursion` method to `DeserializerBuilder`. It allows to limit the maximum allowed depth when deserializing a document. This is particularly useful when parsing untrusted YAML as allowing unbounded depth may lead to a stack overflow which might crash the process.

The signature of the method is the same as the one on `SerializerBuilder`, but in this case there is no default limit as adding one would be a breaking change.

I did reuse the existing `RecursionLevel` class to control the recursion but had to make a few adjustments as I felt it was useful to have the start and end markers in the exception. It was also necessary to add an overload to the SerializerState class to enable constructors with parameters.

I have added a few tests for both methods since the one on `SerializerBuilder` didn't have any.

### DIFF
--- a/YamlDotNet/Serialization/DeserializerBuilder.cs
+++ b/YamlDotNet/Serialization/DeserializerBuilder.cs
@@ -64,6 +64,7 @@ namespace YamlDotNet.Serialization
         private bool enforceNullability;
         private bool caseInsensitivePropertyMatching;
         private bool enforceRequiredProperties;
+        private int? maximumRecursion;
 
         /// <summary>
         /// Initializes a new <see cref="DeserializerBuilder" /> using the default component registrations.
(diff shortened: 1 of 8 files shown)


EdwardCooke:
Also, there's a static deserializer builder now that will need to be updated.

YamlDotNet/Serialization/DeserializerBuilder.cs
```diff
@@ -480,6 +481,24 @@ public DeserializerBuilder WithDuplicateKeyChecking()
             return this;
         }
 
+        /// <summary>
+        /// Sets the maximum recursion that is allowed while building the object graph.
+        /// </summary>
+        /// <remarks>
+        /// Setting this limit is stringly recommended when parsing untrusted input since
```
EdwardCooke: Type-o on stringly. Should probably be strongly.

EdwardCooke:
PR build should be fixed again.

aaubry:
This feature has been released in [version 17.0.0](https://github.com/aaubry/YamlDotNet/releases/tag/v17.0.0).

aaubry:
This feature has been released in [version 17.0.0](https://github.com/aaubry/YamlDotNet/releases/tag/v17.0.0).

PR #1101, opened 2026-04-23
Fix Folded style adds extra newlines 

samuelcadieux:
I'm leaving this PR as a draft for now since this requires modifying the standard yaml test-suite as it stands: https://github.com/yaml/yaml-test-suite/

I suspect I'm not fully understanding the issue at hand.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,5 +3,5 @@
 	url = https://github.com/aaubry/BuildUtils.UnityPrerequisites.git
 [submodule "yaml-test-suite"]
 	path = YamlDotNet.Test/yaml-test-suite
-	url = https://github.com/yaml/yaml-test-suite
+	url = https://github.com/samuelcadieux/yaml-test-suite.git
 	branch = data-2022-01-17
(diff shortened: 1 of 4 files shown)


YamlDotNet.Test/yaml-test-suite
samuelcadieux, line 1: I'm probably misunderstanding what the relationship between in.yaml/out.yaml in this standardized test suite.
Those fixes are just so the tests would pass.

samuelcadieux: From what I gather, those out.yaml are supposed to be the canonical output. It would be equivalent to serializing the rep model (not the output from the emitter).

We probably need the scalar node to store both the raw value and the transformed (folded) value to fix those tests or add those to the ignore test list.

samuelcadieux:
I was cleaning up my implementation so it can do a proper PR, but I think I've just realized what I'm trying to do goes against how YamlDotNet works.

BestWidth made me realize that YamlDotNet regards strings as logical value, and the Emitter takes care of the presentation logic. The only way to have control over the formatting of the string is therefore to use literal strings.

To achieve what I was going for: a pre-formatted fold. I need to send the scalar value with a literal style and pre-process the string applying the folding logic on my end.

Issue #1083, opened 2026-03-23
Reporting a Security Issue


danfiedler-msft:
We'd like to privately report a lower severity vulnerability to YamlDotNet but the project currently lacks a `security.md`. What's the preferred way to report a security issue?

PR #1102, opened 2026-04-27
Security improvements

EdwardCooke:
A couple of vulnerabilities were received.

1. Unbound events could lead to memory exhaustion in the merging parser. This could be a breaking change for large YAML files. It is currently set to 100k events that can be parsed. If you need more, you can set it in the constructor of the merging parser. - Dan Fiedler - Microsoft #1083 
2. YamlDotNet can consume unbounded memory when parsing YAML with many unique keys, anchors, or tags since interning strings are never released - 
Piotr Kiełkowicz - Cisco

Also bumps net8 to net10 on a couple of projects and makes it so net47 isn't ran on Linux environments.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,13 @@
+# Reporting
+
+Please submit reports of security vulnerabilities and, if possible, code to reproduce the vulnerability.
+
+That code will be the basis for the fix.
+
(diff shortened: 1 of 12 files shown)


Kielek:
@EdwardCooke, thanks! Looking for official release.

It will be happy to disclose both issues under https://github.com/aaubry/YamlDotNet/security. It will be easily discoverable by all NuGet users.

FYI: @danfiedler-msft, as you reported first issue.

PR #1107, opened 2026-05-27
Add support for uninitialized instance creation during deserialization

IamFastre:
Add support for uninitialized instance creation during deserialization, mainly to address positional properties records.

* Added `UnstableAttemptUninitializedInstance` setting and related method `EnableUnstableAttemptUninitializedInstance` in `BuilderSkeleton`.
* Now if `DefaultObjectFactory` fails while the option is on it tries `RuntimeHelpers`/`RuntimeSerialization`.
* Added proper tests with positional properties records.

---

So something like:
```cs
public record class Person(string Name, int Age);
```

```cs
    .EnableUnstableAttemptUninitializedInstance()
    .Build();

deserializer.Deserialize<Person>(input);
```

would work.

### DIFF
--- a/YamlDotNet.Test/Serialization/DeserializerTest.cs
+++ b/YamlDotNet.Test/Serialization/DeserializerTest.cs
@@ -768,7 +768,7 @@ year_of_birth: 1983
 car:
   name: Mercedes
   year: 2018";
-            var actual = deserializer.Deserialize<PersonWithCardRecord>(yaml);
+            var actual = deserializer.Deserialize<PersonRecordWithCarRecord>(yaml);
             Assert.Equal("Jack", actual.Name);
             Assert.Equal(1983, actual.YearOfBirth);
             Assert.Equal("Mercedes", actual.Car.Name);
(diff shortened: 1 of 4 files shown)


YamlDotNet/Serialization/ObjectFactories/DefaultObjectFactory.cs
```diff
@@ -109,6 +111,14 @@ public override object Create(Type type)
             {
                 return Activator.CreateInstance(type, settings.AllowPrivateConstructors)!;
             }
+            catch (MissingMethodException) when (settings.UnstableAttemptUninitializedInstance)
+            {
+#if NET5_0_OR_GREATER
```
EdwardCooke, line 116: Lets change this to NET10_0_OR_GREATER.
I'd say NET8 but it's going EOL in a couple of months.

YamlDotNet.Test/Serialization/DeserializerTest.cs
```diff
@@ -696,6 +696,92 @@ public enum EnumMemberedEnum
         }
 #endif
 
+#if NET6_0_OR_GREATER
```
EdwardCooke, line 699: Lets change this to NET10_0_OR_GREATER.
I'd say NET8 but it's going EOL in a couple of months.

YamlDotNet/Serialization/ObjectFactories/DefaultObjectFactory.cs
```diff
@@ -109,6 +111,14 @@ public override object Create(Type type)
             {
                 return Activator.CreateInstance(type, settings.AllowPrivateConstructors)!;
             }
+            catch (MissingMethodException) when (settings.UnstableAttemptUninitializedInstance)
```
EdwardCooke, line 114: Can you find a way to implement this in the staticserializer project as well? I try to keep the 2 with as close to the same featureset as possible.

Issue #1052, opened 2025-04-18
Support for JSON/JSONC Style Comments?


DavidBerg-MSFT:
Most JSON parsers now support comments, and there's a new filetype, [JSONC](https://jsonc.org/), which explicitly supports comments.  To stay compatible with JSON/JSONC implementations, it would be nice if YamlDotNet supported JSON style comments.

Right now, I have a JSON file with some "//" comments that I was parsing with JSON.NET, but now I'm parsing it with YamlDotNet, and had to change the comment to a "#", which is definitely NOT valid JSON or JSONC (so it gets an angry red underline in the editor as the file is still .JSON).

JSON parsers that handle comments include:
* [JSON.Net](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Linq_CommentHandling.htm)
* [System.Text.JSON](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsoncommenthandling?view=net-9.0)

It looks like this would be a fairly straight forward change to [ProcessComment in Scanner.cs](https://github.com/aaubry/YamlDotNet/blob/master/YamlDotNet/Core/Scanner.cs#L651)?

If there's a backwards compatibility concern, I'd be okay with this being under a mode switch.

DavidBerg-MSFT:
I realized this is a broader YAML standard issue, not just related to this product, so I opened [Yaml.Org Issue 196](https://github.com/yaml/www.yaml.org/issues/196) and we're discussing the pros and cons there.

I've also opened [PR 1112](https://github.com/aaubry/YamlDotNet/pull/1112) with a first draft to make it easier to discuss and iterate on implementation.

PR #1112, opened 2026-07-01
Add JSON Comment support

DavidBerg-MSFT:
First attempt at basic JSON comment support.  Currently only supports `//` and not `/*` also causes problems with existing unquoted scalars (like `https://my.org`), which I intend to fix, but I wanted to post what I have so far and get your feedback, so I can incorporate into future updates.

See [Yaml.Org Issue 196](https://github.com/yaml/www.yaml.org/issues/196) for relevant discussion on how best to address the unquoted scalar issue (also [YamlDotNet issue 1052](https://github.com/aaubry/YamlDotNet/issues/1052)).

## Mode Switch

Currently governed by a `Scanner.JsonComments` bool property to turn on/off.  This works for testing, but probably not ideal for plumbing all the way up.  Preferences welcome.

## Unit Tests

I added clones of the existing comment unit tests, but clearly additional tests will be necessary to handle the edge cases (see  [Yaml.Org Issue 196](https://github.com/yaml/www.yaml.org/issues/196)).

I'm seeing failures in the ConformsWithYamlSpec tests, "System.InvalidOperationException : No data found for YamlDotNet.Test.Spec.ParserSpecTests.ConformsWithYamlSpec".  Those failures are occurring with and without my changes, so I'm ignoring them for now; however, if there's something I can do to make those tests work, please let me know.

## Other

I'm generally trying to follow the contribution guidelines, if I missed anything, please let me know.

### DIFF
--- a/YamlDotNet.Test/Core/ScannerTests.cs
+++ b/YamlDotNet.Test/Core/ScannerTests.cs
@@ -390,6 +390,73 @@ namespace YamlDotNet.Test.Core
         }
 
         [Fact]
+        public void JsonCommentsAreReturnedWhenRequested()
+        {
+            var sut = new Scanner(Yaml.ReaderForText(@"
+                    // Top comment
+                    - first // Comment on first item
+                    - second
+                    // Bottom comment
+                "), skipComments: false);
+            sut.JsonComments = true;
+            AssertSequenceOfTokensFrom(sut,
+                StreamStart,
+                StandaloneComment("Top comment"),
+                BlockSequenceStart,
+                BlockEntry,
+                PlainScalar("first"),
+                InlineComment("Comment on first item"),
+                BlockEntry,
+                PlainScalar("second"),
+                StandaloneComment("Bottom comment"),
+                BlockEnd,
+                StreamEnd);
+        }
+
+        [Fact]
+        public void JsonCommentsAreCorrectlyMarked()
+        {
+            var sut = new Scanner(Yaml.ReaderForText(@"
+                - first // Comment on first item
+            "), skipComments: false);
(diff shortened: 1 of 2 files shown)


EdwardCooke:
For those tests, you need to restore the git submodules. That will bring in the conformance tests and then the test problem will go away.

EdwardCooke:
The setting should also be accessible through the deserializerbuilder and staticdeserializerbuilder. Also, can you justify why I should go with your solution over #1117?

DavidBerg-MSFT:
Actually, it looks like Kira-NT has the better solution (#1117<https://github.com/aaubry/YamlDotNet/pull/1117>), supporting both line and multi-line comments, and it lacks the bug I have in processing comments inside identifiers, so I'd go with that one over mine.

Two considerations though:

  1.  He's intentionally not surfaced the option except through an Internal property.  I appreciate that leaves you with the decision to support it; however, if you do want to feature it, I would surface it and not keep it internal.
  2.  There's some general cleanup in my code, such as replacing "while (analyzer.IsWhite()) { Skip(); }" with the existing "SkipWhitespaces();".  It's a small consistency issue, not worth taking my fix over his for, but possibly worth fixing up on its own.

Thanks for following up!

Kira-NT:
>  1.  **He**'s intentionally not surfaced the option except through an Internal property.

Please, do not blindly assume that everyone who writes code is a man. It's never been true.

>  2.  There's some general cleanup in my code, such as replacing "while (analyzer.IsWhite()) { Skip(); }" with the existing "SkipWhitespaces();".

These two approaches are not interchangeable: they have different meanings and produce different results. So this isn't just a "cleanup".

PR #1113, opened 2026-07-08
Add support for implicit string conversions

binerdy:
Allow the serializer to handle objects that use implicit string conversions.


EdwardCooke:
Is this going to mess with the reflection and AoT/trimmed binaries? Reflection usage was abstracted out to another class and the static analyzer implements a concrete, non reflection based version of that class.

binerdy:
I am not sure. If that could pose a risk, it would be good if I could somehow check this. Or at least prove that it would not be the case.

EdwardCooke:
Run a `dotnet publish` in the core7aot test directory. It’ll tell you if there is problems.

binerdy:
Yes, your suspicion was correct.

```
Trim analysis error IL2067: YamlDotNet.ReflectionExtensions.GetPublicStaticMethods(Type): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.NonPublicMethods' in call to 'System.Reflection.RuntimeReflectionExtensions.GetRuntimeMethods(Type)'. The parameter 'type' of method 'YamlDotNet.ReflectionExtensions.GetPublicStaticMethods(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
    D:\packages\.nuget\packages\microsoft.dotnet.ilcompiler\10.0.9\build\Microsoft.NETCore.Native.targets(330,5): error MSB3077: The command ""D:\packages\.nuget\packages\runtime.win-x64.microsoft.dotnet.ilcompiler\10.0.9\tools\\ilc" @"obj\Release\net10.0\win-x64\native\YamlDotNet.Core7AoTCompileTest.ilc.rsp"" exited with return value 0, but errors were detected during execution. ExitCode was set to -1.

Build failed with 2 error(s) in 11,7s
```

binerdy:
Should I look for a different solution, or close this item?

EdwardCooke:
If you want to keep it in the library you’ll need to find a way to go about it without reflection. Or. Move it into the reflection extensions and modify the type inspectors (I think that’s where I did it) and the static serializer to support it.

binerdy:
Moved the implementation into the reflection extensions. To be honest I don't know if the library should support implicit string conversions or not. If it does it probably should support all implicit primitive type conversions... Feel free to reject this PR.

PR #1117, opened 2026-07-20
Add support for JSON comments

### LINKED ISSUE #1052: Support for JSON/JSONC Style Comments?

Kira-NT:
### Description

Hi! This PR adds optional support for JSON-style comments.

YAML 1.2 is, by design, a superset of JSON. Unfortunately, though, with the widespread adoption of formats like JSONC, the ability to use a single YAML parser for both YAML and JSON has become less relevant over time. In my case, for example, I'm working on a tool that I want to keep as small as possible when compiled with NativeAOT to simplify its distribution. However, it needs to parse both YAML and JSON with comments, which forces me to depend on both `System.Text.Json` and `YamlDotNet` - even though I'd happily sacrifice some performance when parsing strictly JSON files if it meant I could trim `System.Text.Json` away.

This PR should help people in similar situations by allowing `YamlDotNet` to process comments according to the [JSONC specification](https://jsonc.org/), if they choose to enable the feature.

### Usage

Compatibility mode for JSON-style comments is enabled by setting `Scanner.AllowJsonComments` to `true`.

You may notice that this property is marked as `internal`. My reasoning is that, since this feature has nothing to do with the YAML specification, it's not really fair to ask the `YamlDotNet` developers to support it. At the same time, making the flag part of the public API would implicitly commit them to doing so. For that reason, I made `AllowJsonComments` `internal`, as this leaves room for the maintainers to modify or even remove the feature entirely if it ever becomes too burdensome, without having to treat the change as a breaking one.

Users who want to opt into this feature can do so via:

```csharp
typeof(Scanner).GetProperty("AllowJsonComments", BindingFlags.NonPublic | BindingFlags.Instance).SetValue(scanner, true);
```

Or, better yet:

```csharp
[UnsafeAccessor(UnsafeAccessorKind.Method, Name = "set_AllowJsonComments")]
static extern void SetAllowJsonComments(Scanner scanner, bool allowJsonComments);

// ...

SetAllowJsonComments(scanner, true);
```

That said, I may be overthinking this a bit, and perhaps simply marking the property with `[EditorBrowsable(EditorBrowsableState.Never)]` would communicate the same intent without it being so overzealous. Let me know what you think!

### Example

When `AllowJsonComments` is enabled, `YamlDotNet` becomes capable of parsing **any** JSONC file *(and even JSONC embedded within YAML)*. Here's an example taken directly from the new tests:

```jsonc
{
    /*
     * Top comment.
     */
    /*0*/ /*1*/"foo"/*2*/ /*3*/:/*4*/ /*5*/false/*6*/ /*7*/,// 8
    // Middle comment.
    /*9*/"bar"/*10*/:/*11*/true//12
    /*
     * Bottom comment.
     */
}
```

By default, it's parsed as:

```json
{
    "/* * Top comment. */ /*0*/ /*1*/\"foo\"/*2*/ /*3*/:/*4*/ /*5*/false/*6*/ /*7*/": null,
    "// 8 // Middle comment. /*9*/\"bar\"/*10*/:/*11*/true//12 /* * Bottom comment. */": null
}
```

With `AllowJsonComments` set to `true`, however, it becomes:

```json
{
    "foo": false,
    "bar": true
}
```

### Implementation

When `AllowJsonComments` is `false` *(i.e., the default)*, the logic of `Scanner` remains completely unaltered.

However, funnily enough, even when I set it to `true` by default and ran the test suite to see which YAML features would break in that mode, all tests still passed. Obviously, tests are imperfect, but hopefully this gives you an idea of how unintrusive this feature is.

As for the notable differences in parsing behavior, consider the following example:

```yaml
{
  foo: https://github.com,
  bar: null/*null*/,
  baz: baz/*baz*/,
  // one: 1,
  // two
}
```

By default, `YamlDotNet` parses it as:

```json
{
    "foo": "https://github.com",
    "bar": "null/*null*/",
    "baz": "baz/*baz*/",
    "// one": "1",
    "// two": null
}
```

With `AllowJsonComments` enabled, it becomes:

```json
{
    "foo": "https://github.com",
    "bar": null,
    "baz": "baz/*baz*/"
}
```

This is definitely not something you would typically see in the real world, so it is unsurprising that the current test suite does not catch these differences. Nevertheless, it is still worth noting that they exist.

### Remaining Questions

- Should `AllowJsonComments` remain `internal`?
- Should its effect be limited to flow collections, given the focus on JSONC compatibility, or should the following example remain valid?
  ```yaml
  // First comment.
  - foo // Second comment.
  ```

---

Closes #1052
Closes #1112


EdwardCooke:
This should be able to be enabled through the deserializerbuilder and staticserializerbuilder.

YamlDotNet/Helpers/JsonHelper.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            // Check whether the value represents a valid number.
+            var i = 0;
+            var hasDot = false;
+            var hasExponent = false;
+
+            // The first character after an optional minus sign must be a digit.
+            // Leading zeros are not allowed.
+            if (s[i] == '-' && ++i == l || (uint)(s[i] - '0') > 9 || s[i] == '0' && ++i < l && (uint)(s[i] - '0') <= 9)
```
EdwardCooke, line 51: What would happen if the there was a malformed json where they put in just a `-`?

YamlDotNet/Helpers/JsonHelper.cs
```diff
@@ -0,0 +1,96 @@
...
+
+using System.Text;
+
+namespace YamlDotNet.Helpers
+{
+    internal static class JsonHelper
+    {
+        public static bool IsJsonLiteral(StringBuilder s)
+        {
+            var l = s.Length;
```
EdwardCooke, line 30: Please use variable names that make more sense as to what they are. For example, this should be length not just `l`.

EdwardCooke: Applies to all of the variable names in this class.

EdwardCooke:
You should definitely fix this use case since it's inside of a quoted string.

`"bar": "null/*null*/",`

EdwardCooke:
You also say you close #1112 which effectively does the same thing. Can you justify why I should go with your implementation over the other one?

Kira-NT:
> This should be able to be enabled through the deserializerbuilder and staticserializerbuilder.

Got it. I just wasn't sure if this feature would be welcome enough here to expose it like that.

> You should definitely fix this use case since it's inside of a quoted string.
>
> `"bar": "null/*null*/",`

Could you elaborate, please? `{ "bar": "null/*null*/", }` is parsed as `{ "bar": "null/*null*/" }`, of course, as there's nothing looking for comments inside string literals. Do you have a complete example that doesn't work for you?

> You also say you close `#1112` which effectively does the same thing. Can you justify why I should go with your implementation over the other one?

It does not do the same thing. It's an unfinished prototype that only handles single-line comments, and only in some cases. This implementation, however, is complete and fully tested.

YamlDotNet/Helpers/JsonHelper.cs
Kira-NT: `-` is not a valid number, so the condition evaluates to `true`, and the method returns `false`.

This exact scenario is covered by `JsonHelperTests.IsJsonLiteralReturnsFalseForNonLiteralValues` on line 94.

EdwardCooke:
I thought I saw in your comments that it didn't return correctly when the value was `null/*null*/`. Looks like I was mistaken. No worries there. But yes, this is a welcome feature so adding it to the builders would be greatly appreciated. Try not to break existing constructors, just add new ones if necessary that call the old ones. That way we don't have to have people do code changes to use the new library.

Your comments are correct on the differences. We'll go with this one and close out the other one.

YamlDotNet.Test/Helpers/JsonHelperTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        public void IsJsonLiteralReturnsTrueForBooleans()
+        {
+            Assert.True(JsonHelper.IsJsonLiteral(new StringBuilder("true")));
+            Assert.True(JsonHelper.IsJsonLiteral(new StringBuilder("false")));
+        }
+
+        [Fact]
+        public void IsJsonLiteralReturnsTrueForNumbers()
+        {
+            Assert.True(JsonHelper.IsJsonLiteral(new StringBuilder("0")));
```
EdwardCooke, line 46: You can pass these in individually as data attributes. It'll be easier to tell what breaks if you do it that way.

EdwardCooke: Heres an example:

https://github.com/aaubry/YamlDotNet/blob/4f7e6e346649ec2a402607a3c1ea352c6fb5ecd3/YamlDotNet.Test/Serialization/DateTimeConverterTests.cs#L65

Kira-NT: Yup, I know. It's just that there are quite a few tests that don't do that, so I just tried to mimic the local style of doing things. Will gladly switch it up otherwise :)

YamlDotNet/Helpers/JsonHelper.cs
```diff
@@ -0,0 +1,96 @@
...
+                    // The number may contain only a single decimal point,
+                    // which must be followed by at least one digit.
+                    if (hasDot || ++i == l || (uint)(s[i] - '0') > 9)
+                    {
+                        return false;
+                    }
+
+                    hasDot = true;
+                }
+                else if ((c | 0x20) == 'e')
```
EdwardCooke, line 70: Whats the reasoning for the `| 0x20` here? Why not just check if `c == 'e'`?

Kira-NT: It's an old "trick" for case-insensitive ANSI comparison. It's a tad faster than `c == 'e' | c == 'E'`, since it boils down to a single comparison, and it's measurably faster than dispatching a whole method call to `char.ToLower(c)`, which also needs to handle proper Unicode runes.

YamlDotNet/Helpers/JsonHelper.cs
Kira-NT: With `l`, I agree. I honestly renamed it from `length` solely to make some of the conditions involving it a bit easier on the eye by making them physically shorter, but it might have affected their readability.

Which other variables would you want me to rename, though? The only ones remaining are `s`, which is just a string we are checking, so any other name like `str`, `input`, etc. wouldn't be any more descriptive; `i`, which should be fine, as it's a universally accepted short name for an index variable; and `c`, which is similar to `s`, just a character we are checking, so naming it something like `chr` or `currentChar` wouldn't make much difference in terms of readability either, at least in my opinion.

Kira-NT:
So, about exposing this feature via `(Static)DeserializerBuilder`: I'm afraid it might cause some confusion. The thing is, `IDeserializer` contains overloads such as `object? Deserialize(IParser parser)`, which might lead to situations like this:

```csharp
var input = "foo: bar // baz";
var deserializer = new DeserializerBuilder().WithAllowJsonComments(true).Build();
var value1 = deserializer.Deserialize(input); // { "foo": "bar" }
var value2 = deserializer.Deserialize(new Parser(new StringReader(input))); // { "foo": "bar // baz" }
```

I.e., even if a deserializer built by `(Static)DeserializerBuilder` is configured to internally create `Parser` instances capable of parsing JSON comments when the user invokes methods like `object? Deserialize(string input)`, `object? Deserialize(TextReader input)`, etc., the user can still provide a parser of their own that does not necessarily support the requested syntax.

Is that ok? Any thoughts on the matter?

Kira-NT:
Maybe instead of exposing it on the builder, it would be better to add a couple of extension methods for `IDeserializer`? This way we could ensure that a properly configured parser is used and that it is exactly what the user expects at the time the method is called.

E.g.,

```csharp
[EditorBrowsable(EditorBrowsableState.Never)]
public static class JsoncDeserializerExtensions
{
    public static T DeserializeJsonc<T>(this IDeserializer deserializer, string input);

    public static T DeserializeJsonc<T>(this IDeserializer deserializer, TextReader input);

    public static object? DeserializeJsonc(this IDeserializer deserializer, string input);

    public static object? DeserializeJsonc(this IDeserializer deserializer, TextReader input);

    public static object? DeserializeJsonc(this IDeserializer deserializer, string input, Type type);

    public static object? DeserializeJsonc(this IDeserializer deserializer, TextReader input, Type type);
}
```

Also, a bit of a spoiler: I have a branch that adds JSON5 support as well. It's only a couple of lines of changes, but it does depend on this PR, hence I haven't submitted it yet. If optional support for parsing JSON5 would also be welcome, we could do something similar for it as well:

```csharp
[EditorBrowsable(EditorBrowsableState.Never)]
public static class Json5DeserializerExtensions
{
    public static T DeserializeJson5<T>(this IDeserializer deserializer, string input);

    public static T DeserializeJson5<T>(this IDeserializer deserializer, TextReader input);

    public static object? DeserializeJson5(this IDeserializer deserializer, string input);

    public static object? DeserializeJson5(this IDeserializer deserializer, TextReader input);

    public static object? DeserializeJson5(this IDeserializer deserializer, string input, Type type);

    public static object? DeserializeJson5(this IDeserializer deserializer, TextReader input, Type type);
}
```

PR #1127, opened 2026-08-07
Serialize types implementing only IReadOnlyDictionary<,> as mappings

### LINKED ISSUE #606: Classes extending IReadonlyDictionary are not serialized as a yaml mapping
We are using a type that extends `IReadOnlyDictionary` and we would like to serialize that type. However, that type is serialized as a collection of key value pairs and not as an object.

I tried seeing about implementing an `IYamlConverter` for `ExtendedReadOnlyDictionary` but it seemed like I would have to re-implement the dictionary conversion.

Is there an extension point in the library that can solve this for me or is an update to the library required?

Example:
The test below returns the following output:
```txt
Assert.AreEqual failed. Expected:<foo: bar
>. Actual:<- Key: foo
  Value: bar
>. 
```

```cs
namespace Testing
{
    [TestClass]
    public class YamlDotNetTests
    {
        [TestMethod]
        public void Serialize_ShouldSerializeDictionaryExtensionsAsObjects()
        {
            // arrange
            IReadOnlyDictionary<string, string> roDict = new Dictionary<string, string>
            {
                {"foo", "bar"}
            };

            ExtendedReadOnlyDictionary<string, string> eroDict = new ExtendedReadOnlyDictionary<string, string>(roDict);

            // act
            var rodYaml = new SerializerBuilder()
                .Build()
                .Serialize(roDict);

            var erodYaml = new SerializerBuilder()
                .Build()
                .Serialize(eroDict);

            // assert
            Assert.AreEqual(rodYaml, erodYaml);
        }

        private class ExtendedReadOnlyDictionary<TKey, TValue> : IReadOnlyDictionary<TKey, TValue>
            where TKey : notnull
        {
            private readonly IReadOnlyDictionary<TKey, TValue> backingDictionary;

            public ExtendedReadOnlyDictionary(IReadOnlyDictionary<TKey, TValue> backingDictionary)
            {
                this.backingDictionary = backingDictionary;
            }

            public TValue this[TKey key] => this.backingDictionary[key];

            public IEnumerable<TKey> Keys => this.backingDictionary.Keys;

            public IEnumerable<TValue> Values => this.backingDictionary.Values;
            public int Count => throw new NotImplementedException();

            public bool ContainsKey(TKey key)
            {
                return this.backingDictionary.ContainsKey(key);
            }

            public IEnumerator<KeyValuePair<TKey, TValue>> GetEnumerator()
            {
                return this.backingDictionary.GetEnumerator();
            }

            public bool TryGetValue(TKey key, [MaybeNullWhen(false)] out TValue value)
            {
                var result = this.backingDictionary.TryGetValue(key, out var foundValue);

                value = foundValue;

                return result;
            }

            IEnumerator IEnumerable.GetEnumerator()
            {
                return this.backingDictionary.GetEnumerator();
            }
        }
    }
}
```

youdie006:
## Motivation

Fixes #606. A type that implements only `IReadOnlyDictionary<TKey, TValue>` (and neither the non-generic `IDictionary` nor `IDictionary<TKey, TValue>`) is serialized as a **sequence of `{ Key, Value }` mappings** instead of a single YAML **mapping**:

```yaml
# actual (before)
- Key: key1
  Value: value1
- Key: key2
  Value: value2

# expected
key1: value1
key2: value2
```

`FullObjectGraphTraversalStrategy.TraverseObject` dispatches dictionaries via `ObjectFactory.GetDictionary`, which only recognises `IDictionary<,>`. A read-only-only dictionary therefore returns `false` there and falls through to the `IEnumerable` branch (`IReadOnlyDictionary<,>` is `IEnumerable<KeyValuePair<,>>`), so it is written as a list. The `BCL ReadOnlyDictionary<,>` happens to also implement `IDictionary`, so only custom read-only dictionary types are affected. The asymmetry is real: the deserialize side already handles non-`IDictionary` dictionary types (see the existing `GenericDictionaryThatDoesNotImplementIDictionaryCanBeDeserialized` test).

## Solution

- `ObjectFactoryBase.GetDictionary` now also detects `IReadOnlyDictionary<,>` (after the existing `IDictionary<,>` check, so types implementing both keep the current path) and wraps the value in a new `GenericReadOnlyDictionaryToNonGenericAdapter<TKey, TValue>`.
- The new adapter mirrors the existing `GenericDictionaryToNonGenericAdapter` but takes an `IReadOnlyDictionary<,>`. Object-graph traversal (`TraverseDictionary`) only enumerates the adapter, so only `GetEnumerator` is implemented; the rest throw `NotSupportedException`, exactly like the existing adapter's unused members.

The regular `IDictionary<,>` path is unchanged.

## Test

Added `SerializeGenericReadOnlyDictionary`, plus a `GenericTestReadOnlyDictionary<,>` test helper (implements only `IReadOnlyDictionary<,>`, mirroring the existing `GenericTestDictionary`). It asserts the read-only dictionary serializes to the same YAML as an equivalent `Dictionary<,>`.

- Before the fix: fails (serializes as a sequence).
- After the fix: passes.

Verified locally in Docker: `dotnet test` — the full `YamlDotNet.Test` suite passes (the only 2 failures, `ConformsWithYamlSpec`, are pre-existing and caused by the YAML spec test-suite submodule not being present in the shallow checkout; they fail identically without this change). The library also builds cleanly for `netstandard2.0` and `net8.0` (0 warnings).

---

Disclosure: this fix was prepared with AI assistance (Claude). I reviewed it, confirmed the root cause in the traversal dispatch, and verified the red-green test and full suite myself.


EdwardCooke:
Can you make this work with the static object factory project as well? I try to keep both the default reflection based implementation and the static one working as well for those that need the AoT support.

